PR #39, opened 2020-03-09
Changed localhost to correct host

rrikooo:


Kaltsoon:
Thanks, we will fix this!

PR #163, opened 2020-03-19
Fix link to intersection type documentation in part 9d

TamarixGallica:
Original link leads to 404. Intended documentation seems to be available at https://www.typescriptlang.org/docs/handbook/advanced-types.html#intersection-types.


TamarixGallica:
However, on further reading it seems PropsWithChildren is a union instead of an intersection.

mluukkai:
Thanks. TypeScript team has tested their new documentation pages during weekend and some links have had changes...

PR #170, opened 2020-03-21
Update part6a.md

Prashanga:
In Dan Abramov's Medium article about presentational and container components which is linked in this part in the 'More Components' section, he has an updated edit about not splitting components such way. "This text is left intact for historical reasons but don’t take it too seriously" is on the update.


Prashanga:
Dan's article: https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0

PR #189, opened 2020-03-26
[Incorrect- please delete] Correcting the install bcrypt command

slowdata:
Starting exercise 4.15: bloglist expansion, step4
it stats advising to save encrypted password and as that fact adds the command to install the bcryptjs package. The command uses uninstall as for uninstalling the package.
It should be installing or i.


mluukkai:
this is right, it advices first to _uninstall_ bcrypt and then to install the js-version

PR #192, opened 2020-03-28
build: update deps

thetric:
Updates the dependencies to the latest version. This enables Node 12 compatibility and fixes build errors under Node 10:

```
  Error: Input file is missing or of an unsupported image format

error Failed to process image /home/thetric/github/fullstack-site/src/content/images/5/39ea.png
```


src/components/Element/Element.js
```diff
@@ -19,7 +19,7 @@ const Element = ({
   noWrap,
   centeredInDesktop,
   relative,
-  ...props,
+  ...props
```
thetric: I removed the semicolon to fix the following build error:

> 
>  ERROR #98123  WEBPACK
> 
> Generating development JavaScript bundle failed
> 
> /home/thetric/github/fullstackopen/src/components/Element/Element.js: Rest element must be last element (22:10)
> 
>   20 |   centeredInDesktop,
>   21 |   relative,
> > 22 |   ...props,
>      |           ^
>   23 | }) => {
>   24 |   const classes = [];
>   25 | 
> 
> File: src/components/Element/Element.js:22:10

mluukkai:
the current works with node 10.18, will not merge this because of too much risk

PR #193, opened 2020-03-28
correct syntax highlighting

thetric:


mluukkai:
formatting with ts does not work, looks like this

image

thetric:
This PR requires the updated dependencies from #192 to work (left: this PR, right: current):

image

PR #207, opened 2020-04-01
Fix exercise to specify when to add 'React' part

lhiginbotham:
Modify part 2.2 to tell the student they need to add the 'React' part to the course.  In addition, give better direction on how to refactor and use the 'Total' component

### DIFF
--- a/src/content/2/en/part2a.md
+++ b/src/content/2/en/part2a.md
@@ -691,7 +691,46 @@ Ensure that the console shows no errors!
 <h4>2.2: Course contents step7</h4>
 
 
-Show also the sum of the exercises of the course. 
+Add the "Redux" part to the course:
+
+```js
+const App = () => {
+  const course = {
+    id: 1,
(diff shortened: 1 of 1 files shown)


src/content/2/en/part2a.md
```diff
@@ -691,7 +691,46 @@ Ensure that the console shows no errors!
...
+
+  return (
+    <div>
+      <Course course={course} />
+    </div>
+  )
+}
+```
+
+After this, refactor the <i>Total</i> component to handle rendering the total number of exercises for a variable length collection of course parts.  Update the <i>Course</i> component to utilize the revised <i>Total</i> component, producing an output similar to the following:
```
lhiginbotham, line 733: It seems like this might be overlapping with exercise 2.3, but if the student for some reason chooses to solve 2.2 using a for-loop, then suggesting `reduce` in 2.3 and forcing them to utilize it will be good for setting them back on course, reinforcing good JavaScript practices.

PR #209, opened 2020-04-01
Multiple default export

hugosjoberg:
**WARNING** I am completely new to javascript so this might be a incorrect commit. However, I was not able to import the functions when exporting as suggested in the course material but it worked then applying the changed I have propsed in this commit.


mluukkai:
The code is correct, please check that you are importing it as follows

https://github.com/fullstack-hy2020/part2-notes/blob/part2-6/src/App.js#L4

PR #221, opened 2020-04-05
Add warning about json modules to en/part9c

jeremy-ebinum:
While doing exercise 9.16 I ran into a problem extending a typescript interface with a new property.

I had a flat directory "data" with contents:

├── patients.json
└── patients.ts

and tried to import the patients.ts file but I hadn't realized earlier that because I set _resolveJsonModule_ to true in _tsconfig.json_, _patients.json_ would now be included in node module resolution.

So while trying to:

`import patients from "./data/patients"`

Typescript kept compiling the _patients.json_ file instead of _patients.ts_ so my new object with supposed new properties was using stale data from the _.json_ file and with type assertions it was difficult to track down the exact problem.

I made an edit to the English version of part9c that includes a warning about this behavior.

### DIFF
--- a/src/content/9/en/part9c.md
+++ b/src/content/9/en/part9c.md
@@ -603,6 +603,57 @@ export interface DiaryEntry {
 } 
 ```
 
+### Node and JSON modules
+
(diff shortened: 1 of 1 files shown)


mluukkai:
👍

PR #231, opened 2020-04-05
Change "npx" to "npm"

callvm:


callvm:
Whoops, my mistake. Npx is right

Issue #246, opened 2020-04-09
Mongoose 5.9.7 Breaking Error Handler Middleware (Part 3c onwards)


ScheuNZ:
Mongoose 5.9.7 is breaking the error handler middleware and causing the "invalid id" unit test to fail because the middleware no longer recognises that the `CastError` should yield a 400 HTTP response. 

More specifically when the "invalid id" is sent to the database during the unit test, version 5.9.7 of the Mongoose library is raising an exception where the `kind` property of the error is `undefined` instead of `ObjectId`. This in turn is causing the error handler middleware to fall though and yield a 500 response instead of the 400 the unit test expects.

```
// Error handler middleware from utils\middleware.js
const errorHandler = (error, request, response, next) => {
  // The if statement below is broken in Mongoose 5.9.7 because error.kind is undefined
  if (error.name === 'CastError' && error.kind === 'ObjectId') {
    return response.status(400).send({ error: 'malformatted id' })
  }
```

An example of the exception from Mongoose 5.9.7 is shown below:

```
  console.error utils/logger.js:8
    Cast to ObjectId failed for value "5a3d5da59070081a82a3445" at path "_id" for model "Note"

  console.log utils/middleware.js:18
    error: MongooseError {
      message: 'Cast to ObjectId failed for value "5a3d5da59070081a82a3445" at path "_id" for model "Note"',
      name: 'CastError',
      messageFormat: undefined,
      stringValue: '"5a3d5da59070081a82a3445"',
      kind: undefined,
      value: '5a3d5da59070081a82a3445',
      path: '_id',
      reason: Error: Argument passed in must be a single String of 12 bytes or a string of 24 hex characters
```

I've tested Mongoose 5.9.6 and earlier versions and they do not appear to be affected. I only noticed this because `npm install --save mongoose` currently installs 5.9.7 by default.

The [changelog](https://github.com/Automattic/mongoose/commit/4015625b33affd7e277cf0ff752cedf729fef41b) for the 5.9.7 Mongoose release would indicate cast related functionality has been changed but I have been unable to ascertain whether this `CastError` behaviour change was intentional.

I'm happy to submit a PR for a note in the course somewhere if someone can advise how this problem should be addressed. I imagine more people will be hitting this issue soon if `npm` is installing 5.9.7 Mongoose by default.

boonepeter:
[...]

PR #284, opened 2020-04-14
remove error.kind from error checking

boonepeter:
I have remove the `error.kind` check because it seems that mongoose now returns an error that doesn't have a `kind` defined. 

See this issue: #246, and this other PR I made: #270


mluukkai:
👍

PR #372, opened 2020-05-05
Update part1b.md

takecare:
Proposing some changes to make the language and some examples clearer. Namely:

- Adding jsbin as a suggestion (link displays only editor + console)
- Fixing "e.g." being misplaced in some sentences
- Reworking the "this" examples so they are clearer
- Fixing "this" on setTimeout() referring to the global object (instead of the Timeout object)

### DIFF
--- a/src/content/1/en/part1b.md
+++ b/src/content/1/en/part1b.md
@@ -20,11 +20,11 @@ Today, the most popular way to do the transpiling is using [Babel](https://babel
 
 The code is written into files ending with <i>.js</i> and are run by issuing the command <em>node name\_of\_file.js</em>
 
-It is also possible to write JavaScript code into the Node.js console, which is opened by typing _node_ in the command-line, as well as into the browser's developer tool console. The newest revisions of Chrome handle the newer features of JavaScript [pretty well](http://kangax.github.io/compat-table/es2016plus/) without transpiling the code.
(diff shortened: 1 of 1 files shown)


src/content/1/en/part1b.md
```diff
@@ -386,12 +386,12 @@ const arto = {
   education: 'PhD',
   // highlight-start
   greet: function() {
-    console.log('hello, my name is', this.name)
+    console.log('hello, my name is ' + this.name)
```
takecare, line 389: imo this example is easier to understand for beginners this way, as opposed to logging `this.name` as a second param to `console.log`, as the result that we see in the comments is the actual result the user will see if they try it out for themselves

src/content/1/en/part1b.md
```diff
@@ -440,43 +440,43 @@ const referenceToAddition = arto.doAddition
...
 
 If we try to do the same with the method _greet_ we run into an issue:
 
 ```js
-arto.greet()       // hello, my name is Arto Hellas gets printed
+arto.greet()       // "hello, my name is Arto Hellas" gets printed
 
 const referenceToGreet = arto.greet
-referenceToGreet() // prints only hello, my name is
+referenceToGreet() // prints "hello, my name is undefined"
```
takecare, line 451: here in particular is the place where i think this concatenation in the example is advantageous. before, the comment is not really helpful in showing what happens.

PR #379, opened 2020-05-06
Update part1d.md

takecare:
Proposing some changes to make the language and some examples clearer. Namely:

- Fixing the usage of `-` throughout this part
- Changing some of the language to avoid repetition

### DIFF
--- a/src/content/1/en/part1d.md
+++ b/src/content/1/en/part1d.md
@@ -128,7 +128,7 @@ const handleRightClick = () => {
 }
 ```
 
-The syntax may seem a bit strange at first. In practice <em>{ ...clicks }</em> creates a new object that has copies of all of the properties of the _clicks_ object. When we add new properties to the object, e.g. <em>{ ...clicks, right: 1 }</em>, the value of the _right_ property in the new object will be 1.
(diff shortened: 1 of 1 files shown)


src/content/1/en/part1d.md
```diff
@@ -290,7 +290,7 @@ const App = (props) => {
 }
 ```
 
-Now the behavior of the component depends on whether or not any buttons have been clicked. If not, meaning that the <em>allClicks</em> array is empty, the component renders a div component with some instructions:
+Now the behavior of the component depends on whether or not any buttons have been clicked. If not, meaning that the <em>allClicks</em> array is empty, the component renders a div element with some instructions instead:
```
takecare, line 293: it's actually a component so not sure about this change, but i feel not repeating "component" helps make things easier to understand

src/content/1/en/part1d.md
```diff
@@ -1016,9 +1016,8 @@ The internet is full of React-related material. However, we use such a new style
 
 You may find the following links useful:
 
-- The React [official documentation](https://reactjs.org/docs/hello-world.html) is worth checking out at some point, although most of it will become relevant only later on in the course. Also, everything related to Class-components is irrelevant to us.
-- Beware the official React [tutorial](https://reactjs.org/tutorial/tutorial.html), it's not very good.
```
takecare, line 1020: imo there's no value in stating that this tutorial is not good without explaining why

PR #395, opened 2020-05-11
Removes redundant .toJSON() call

shafkathullah:
I am not sure, but the returned database object is void of _id & -__v fields without calling .toJSON().


Kaltsoon:
Thanks for the pull request! Yes, that's absolutely true. We fixed this and included a short explanation that the `toJSON` method is called automatically.

PR #419, opened 2020-05-17
Update part3b.md

Reyno851:
At line 144, copying production build from front-end to back-end using the command line was only mentioned for MacOS. I propose mentioning the use of the copy or xcopy command to guide Windows users, or simply doing a copy and paste to avoid confusion.


Kaltsoon:
Good stuff. Made some minor changes to your text

Reyno851:
Thanks!

PR #443, opened 2020-05-26
Part9b: The horrors of any -- italicized 'any'

liamst19:
The *any* type keyword should be italicized for the sentence: "Programmers however see the code differently when any is explicitly enforced than when it implicitly inferred."


Kaltsoon:
Thanks!

PR #448, opened 2020-05-27
ENV variables not working if not string

batrashi:
env variables need to be string, at lees for windows powershell.
Otherwise not working with an error

### DIFF
--- a/src/content/2/en/part2c.md
+++ b/src/content/2/en/part2c.md
@@ -643,9 +643,9 @@ Add to the view showing the data of a single country, the weather report for the
 Assuming the api-key is <i>t0p53cr3t4p1k3yv4lu3</i>, when the application is started like so:
 
 ```bash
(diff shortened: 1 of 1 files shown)


Kaltsoon:
That's not the case for Linux/macOS bash at least, but we need to test this with a Windows machine

antonvsdata:
For me, on Windows, the Powershell approach works fine, thank you for finding this!

For some reason, cmd.exe gives me weird strings when I try to pass an environment variable. With the approach in this pull request, the string looks like this: "'e2bfee93633491c03b2c93563324170a' ".
I tried running the command with double quotes and without quotes, but ran into similar problems. This can of course be solved by modifying the string in Node, but I think it is ok to go with the Powershell approach, as I think that is the shell most people use on Windows, and also the default terminal that opens in VS Code.

Kaltsoon:
Awesome, thanks for you contribution!

PR #458, opened 2020-05-30
Fix the use of "comprise of" -> "comprise"

zegheim:
First of all thank you so much for producing this course! It's been great and I'm enjoying it at the moment. 

With that said, I would like to request that the use of the phrase "comprise of" be changed to "comprise" instead. The word comprise means "consists of", so the extra of should not be there.


Kaltsoon:
Thanks for the feedback and for the pull request!

PR #464, opened 2020-06-03
Update part 5b

takecare:
`createRef` is being incorrectly used, so it's best if we replace it with the corresponding hook: `useRef`


Kaltsoon:
This is great, thanks a lot!

PR #467, opened 2020-06-05
Update part4d.md

Nirmal201:
Full Token Required. Eg:- eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6Im5pcm1hbCIsImlkIjoiNWVkNzhjYTM2NDU1ODU0YWNmMDNkMWQ3IiwiaWF0IjoxNTkxMzU0Nzk4fQ.mMaOubp-qmwGvITb-wOaey-8UNNlhlVhBMnTjrqBpjs


Kaltsoon:
That is a bit confusing. If the token is `X` then the format is `Bearer X`. Whether the token in the example is a valid JWT is a different story, but that is not the point here

PR #468, opened 2020-06-06
Keep content consistent with earlier parts

zegheim:
see [Project structure](https://fullstackopen.com/en/part4/structure_of_backend_application_introduction_to_testing#project-structure) 
Ctrl + F "The contents of the notes.js module are the following:" and then look at the code block immediately below it


Kaltsoon:
Thanks!

PR #469, opened 2020-06-07
Part9d Exercise 9.14: fixed broken link

liamst19:
The link `/en/part1/javascript#exercises-1-3-1-5` lead to 404 Not Found page. The actual page was  `/en/part1/java_script#exercises-1-3-1-5`.
This fixes the broken link, but since the word *javascript* has no space, the wrong link seems to be more appropriate than the path of the actually existing one..


Kaltsoon:
Thanks!

PR #470, opened 2020-06-07
Part 8b: Irrelevant 'async'

heyDante:
The `async` keyword is irrelevant since there's no use of `await` inside the mentioned block.


Kaltsoon:
Thanks!

PR #471, opened 2020-06-08
Added additional resources for publishing website.

Mahmud031:
We are publishing the website by using Heroku. There's a lot of other ways for publishing a website and MDN web docs include an introduction for other ways. It can be a useful to have this information.


Kaltsoon:
That is true, however, the MDN documentation is quite vague being more about how to publish a simple website with a index.html file. A resource for deploying a SPA application with a build step etc would perhaps make more sense, but maybe the Heroku deployment is enough for this course

PR #477, opened 2020-06-11
Update part7d.md - Fix path

dekisr:
Change from `'http://localhost:3001/api/notes'` to `'http://localhost:3001/notes'`. Alternatively we could add a `routes.json` file, redirecting to "/api". And then running `json-server --watch db.json --port 3001 --routes routes.json`. So, I think the best approach is just fixing the path.


Kaltsoon:
Yes, this is the case for the json-server. However, the express server has the `/api` prefix in the path: https://fullstackopen.com/en/part3/node_js_and_express

PR #480, opened 2020-06-12
Fix a link on Part0b

bookyue:
According to English version, add a link which should be there for "示例应用" in Chinese version part0b


Kaltsoon:
Thanks!

PR #487, opened 2020-06-14
Update part3d - Added a note in 3.20

yetian:
Added a note w.r.t. name validation in production


Kaltsoon:
A good point, but it isn't that relevant in the scope of this exercise and would most likely just cause confusion

PR #495, opened 2020-06-18
fix link

emilianomacht:
There was a link pointing to "https://fullstackopen.com/em/part8" which results in 404 status. The correct link is "https://fullstackopen.com/en/part8".


Kaltsoon:
Thanks!

PR #497, opened 2020-06-19
Update part3c.md

YBCS:
line 318  has different number as compared to when it is printed in line 332


Kaltsoon:
Thanks!

YBCS:
thank you for making this amazing course open source. I look forward to completing this course and also make more contibutions.

PR #499, opened 2020-06-21
Update Finnish part 8b.

jyraukol:
The second parameter for useEffect is result not result.data. Add information of if statement checking result.data.


Kaltsoon:
Thanks!

PR #502, opened 2020-06-23
Update "no-console" part

shafkathullah:
'no-console' is not part of 'eslint:recommended' rules.


shafkathullah:
The course content contains this line "**...This(ie, `eslint:recommended` config) includes a rule that warns about console.log commands**". While actually it doesn't. There is no harm in turning it off, but I thought the line might cause confusion.

PR #504, opened 2020-06-26
Update osa8d.md

jyraukol:
Object from Apollo store.readQuery is immutable so we can't push to dataInStore.allPersons. Instead set a new object as data.
Add useApolloClient import for clarification in code snippet.
Add useEffect for reading token from localStorage in App.js


Kaltsoon:
True, we should do it similarly as in the English version: https://fullstackopen.com/en/part8/login_and_updating_the_cache#updating-cache-revisited.

Kaltsoon:
The token related code in the `useEffect` doesn't seem to be that relevant

PR #505, opened 2020-06-27
Update part3b - added npm set script-shell to enable bash on windows

JS64:
This requires the default Git for Windows installation which is already a prerequisite for this course.


Kaltsoon:
Good stuff, thanks!

PR #512, opened 2020-07-01
Replaced placeholder with code

antonvsdata:
Line 325: A code snippet only contained a placeholder "(koodi tähän)". It is now replaced with the appropriate code.


Kaltsoon:
Thanks, good catch!

PR #514, opened 2020-07-02
Part 2c: Another way to secure API key

antonvsdata:
It seems that Windows users can have trouble with the environment variables. I tried many different things from StackOverflow, but nothing seemed to work. I think it might be a good idea to add another way of working with the API keys, so that the problems with environment variables do not cause students to avoid the exercise itself. I think saving all sensitive information in a separate config file and not committing that file is a good practice. But I also understand if you feel like environment variables are the way to go.

### DIFF
--- a/src/content/2/fi/osa2c.md
+++ b/src/content/2/fi/osa2c.md
@@ -646,4 +646,14 @@ const api_key = process.env.REACT_APP_API_KEY
 // muuttujassa api_key on nyt käynnistyksessä annettu api-avaimen arvo
 ```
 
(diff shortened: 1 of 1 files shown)


Kaltsoon:
This approach is a bit dirty. Importing a file included in the `.gitignore` isn't a good idea. I would much rather find a solution to use environment variables in Windows successfully. If you have the time, could you check if the solutions suggested in this pull request works on Windows: https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/pull/448/files

antonvsdata:
Hi, I got the environment variables working on Windows with those tips! I think that is the way to go. I'll post more details to the linked pull request.

Kaltsoon:
Awesome, thanks!

PR #519, opened 2020-07-03
Update part3a.md

daniel0707:
Prefer 'dev' over 'tiny' for morgan config, as it color codes logs.


Kaltsoon:
I think we should stick with 'tiny' because 'dev' doesn't make that much difference. We would also need to change the format depending on the environment because 'dev' should be only used in the development environment: 

> Concise output colored by response status for **development use**

PR #535, opened 2020-07-09
Update vulnerable packages and enough to compile on node 14.5.0

vvilhonen:
Ran `npm update` and `npm audit fix` to update packages. Got project to compile on node.js 14.5.0 again with that.


Kaltsoon:
Thanks!

PR #541, opened 2020-07-11
Modified Chinese Content According to the En Version(week28 @ 2020)

RichardStark:


Kaltsoon:
Awesome, thanks a lot!

PR #552, opened 2020-07-16
Update part 3d, fix typo

fmercille:
Fixed typo ("can not" --> "cannot")


Kaltsoon:
Both seem to be correct: https://english.stackexchange.com/a/204010

PR #554, opened 2020-07-16
Add one parenthesis to code example

grekulanssi:
as the code snippet begins Togglable definition with `React.forwardRef((props, ref) => {`, it should also end with `})` instead of only `}`.


Kaltsoon:
Thanks!

PR #556, opened 2020-07-16
5c one more highlight line

grekulanssi:
this line is also new to the student as well as relevant and thus should be highlighted. I also grouped those two lines together.


Kaltsoon:
Thanks!

PR #557, opened 2020-07-16
added date property to test notes in 4b

zakuraevs:
In the 'Initializing the database before tests' section,

the test notes lack "date" property and the tests therefore fail.

### DIFF
--- a/src/content/4/en/part4b.md
+++ b/src/content/4/en/part4b.md
@@ -285,10 +285,12 @@ const initialNotes = [
   {
     content: 'HTML is easy',
     important: false,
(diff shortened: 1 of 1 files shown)


jtarvainen:
The inclusion of the "date" property in the notes also causes the tests to fail, because the note returned by the express router's response.json(...) method has quotes around it, but the note returned by the notesInDb method does not.

Currently, the Finnish-language course page includes the "date" property in the initialNotes definition, but the English-language page does not.

Kaltsoon:
Thanks, this issue has been fixed!

PR #561, opened 2020-07-17
Update part1d.md

ansible-comme:
Moved the sentence start to the right place


Kaltsoon:
I think it makes sense to have the "If the component" before the code rather than after it so that it is clear what "The component" is referring to.

PR #562, opened 2020-07-17
Update part4d.md

waveling:
Just minor spelling errors, but might be confusing to some. At least most scientific sources use "scheme" instead of "schema".


Kaltsoon:
Thanks!

PR #565, opened 2020-07-19
Update part2c.md

ramlo:
Add a comment to indicate that if you add a .env file, you need to restart the server


Kaltsoon:
Thanks!

PR #566, opened 2020-07-20
Remove misleading pointer cursor on sidebar items

vvilhonen:
Navigation sidebar on the left side shows up pointer cursor when hovering on items indicating a clickable region. If the mouse pointer isn't located on top of the anchor element inside of the list item, nothing happens when a mouse button is clicked.

This PR removes the misleading indicator for clickable elements on navigation when there are no anchors under cursor.


Kaltsoon:
Thanks!

PR #577, opened 2020-07-23
Mixup between createNote and createNew

Hakizu:
createNew is a variable in noteServices handling POST methods
createNote is the action creator that is being referred to


Kaltsoon:
Thanks!

PR #580, opened 2020-07-24
converting Date objects to ISO string

nathaniades:
The original notes array contains Date objects that are strings, and the method new Date().toISOString() was used in a previous part.

### DIFF
--- a/src/content/3/en/part3a.md
+++ b/src/content/3/en/part3a.md
@@ -855,7 +855,7 @@ app.post('/api/notes', (request, response) => {
   const note = {
     content: body.content,
     important: body.important || false,
-    date: new Date(),
+    date: new Date().toISOString(),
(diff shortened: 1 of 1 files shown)


Kaltsoon:
How is this necessary, when the default JSON presentation of `Date` object is `Date.toISOString`? In other words: `JSON.stringify({ date: new Date() }) === JSON.stringify({ date: new Date().toISOString() })`

nathaniades:
https://fullstackopen.com/en/part2/forms -> coding style uses toISOString
https://fullstackopen.com/en/part2/altering_data_in_server -> coding style uses toISOString

It's not wrong, just confusing. Why bother include it in some parts of your program but decide to rely on the default implementation on other parts? Either exclude it everywhere or include it everywhere. It's about consistency and clarity.

PR #585, opened 2020-07-27
Update part2a.md

di-su:


di-su:
I think it's a typo, I don't understand why "Part" is repeated

Kaltsoon:
It's not a typo, it just means that there are multiple Part components inside the Content component.

di-su:
> It's not a typo, it just means that there are multiple Part components inside the Content component.

Ah! I don't understand this, I thought there was only one Part. Do you have a solution you can show me?

PR #587, opened 2020-07-28
Fix broken link on part9b.md

zeagel:
Link '[tsconfig page](https://www.typescriptlang.org/v2/en/tsconfig)' seems to be broken. If the aim is that this link should point to TS v2 TSConfig documentation, then the correct link should be perhaps this instead: 'https://www.staging-typescript.org/tsconfig'. Otherwise, if the link should point to the latest TSConfig documentation, then the link should be this: 'https://www.typescriptlang.org/docs/handbook/tsconfig-json.html.


Kaltsoon:
Thanks! The https://www.staging-typescript.org/tsconfig seems to be the most comprehensive

PR #596, opened 2020-07-30
propose: use body parser in Exercise 4.2 step 2 

Shivam-Chahar:
Without the use of a body parser in Exercise 4.2 step 2, incoming JSON object from the post request made with the postman, the whole object body is not getting saved in MongoDB collection. Only the _id produced by MongoDB is getting saved. Using a body-parser will solve this problem

### DIFF
--- a/src/content/4/en/part4a.md
+++ b/src/content/4/en/part4a.md
@@ -413,6 +413,7 @@ Refactor the application into separate modules as shown earlier in this part of 
 
 **NB** refactor your application in baby steps and verify that the application works after every change you make. If you try to take a "shortcut" by refactoring many things at once, then [Murphy's law](https://en.wikipedia.org/wiki/Murphy%27s_law) will kick in and it is almost certain that something will break in your application. The "shortcut" will end up taking more time than moving forward slowly and systematically.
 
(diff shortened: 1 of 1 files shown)


Kaltsoon:
Is this necessary because body-parser middleware is already used in exercise 4.1?

Shivam-Chahar:
It is just necessary to tell which to use and where otherwise beginners will find difficulty in knowing what's happening and why

Kaltsoon:
What do you mean by "which to use"? Doesn't course only use the `express.json` body parser?

Shivam-Chahar:
Nowhere it is mentioned before Exercise 4.2 to use express.json body-parser or even about the body-parser

Kaltsoon:
This is from part 3a:
Screenshot 2020-07-31 at 11 11 10

In addition, as I mentioned exercise 4.1 uses the `express.json` body parser.

Shivam-Chahar:
You are right, But when Express 4.0 was released bundled middleware was removed from Express and made them separate packages instead. The syntax then changed from app.use(express.json()) to app.use(bodyParser.json()) after installing the bodyParser module.

Kaltsoon:
As far as I know, express (>=4.16.0) doesn't encourage to use the `body-parser` package anymore and use `express.json` instead. I think we should follow this guideline and not introduce the `body-parser` package. Supporting older express versions doesn't make sense.

Shivam-Chahar:
okay

PR #597, opened 2020-07-30
Additional note about and error in part9c.md

zeagel:
In the section 'Adding a new diary' in part9c.md, when updating router functionality, it would be wise to add proposed additional note (or such) about an error 'Unsafe assigment of an any value.' which will happen due the line where is done destructuring assigment of request data (which has not been typed yet).


mluukkai:
this is now fixed

PR #598, opened 2020-07-30
Note about unhandled ESlint warnings in part9c.md

zeagel:
At the end of part9c.md, it would be wise to add some disclaimer/additional note about unhandled ESlint warnings. In order to avoid any unnecessary doubts 'Why I get these warnings?! Did I miss something...?' from students. ;-)


mluukkai:
this is now finally fixed

PR #601, opened 2020-07-31
Update part3a.md

orvalho:
By convention, the constant identifiers should be in uppercase. Especially when this convention is used a bit further down the material.


Kaltsoon:
Thanks!

PR #603, opened 2020-08-01
Which's does not exist

Hakizu:
Use who's or whose - depending on the context 
Which's however does not exist as a word.


Kaltsoon:
Thanks!

PR #606, opened 2020-08-03
Fixes the important prop for new notes

nathaniades:
Given that Math.random() results in range of 0 to less than 1, Math.random() < 0.5 is the correct implementation for 50% true/false chance. 
Related https://fullstackopen.com/en/part2/forms and https://stackoverflow.com/questions/44651537/correct-function-using-math-random-to-get-50-50-chance/44651588


Kaltsoon:
Thanks!

PR #624, opened 2020-08-09
Update part2c.md

orvalho:
By convention, react-dom should be imported after importing react and not before. Especially when this convention is already used in the material.


Kaltsoon:
Thanks!

PR #626, opened 2020-08-10
Update part3a.md

orvalho:
It says that curl was mentioned briefly in the previous part of the material. However, it was not.


Kaltsoon:
Thanks!

PR #628, opened 2020-08-10
Remove highlight for notes in userSchema

baciucristian:
Remove highlight because the code it's already in the previous example.


Kaltsoon:
Thanks!

PR #635, opened 2020-08-12
Update part4a.md

orvalho:
PORT and MONGODB_URI are unchanging variables, therefore, const should be used instead of let


Kaltsoon:
That's true, thanks!

PR #643, opened 2020-08-14
Update part4b.md

orvalho:
Original code (line 230) was changed (line 322). The changed code should be highlighted.


Kaltsoon:
Thanks!

PR #651, opened 2020-08-17
Patch 6

lelilia:
Since we added required to the date in the notes module we need to have a date set in the database.

So I added dates to the initial database (because we create the notes with save and not send so the controller for post is not called that adds the date for us

I also added a bit of code to change the date from an object to a string in the model

### DIFF
--- a/src/content/4/en/part4b.md
+++ b/src/content/4/en/part4b.md
@@ -284,10 +284,12 @@ const Note = require('../models/note')
 const initialNotes = [
   {
     content: 'HTML is easy',
+    date: new Date(),
     important: false,
   },
   {
     content: 'Browser can execute only Javascript',
+    date: new Date(),
     important: true,
   },
 ]
--- a/src/content/4/en/part4b.md
+++ b/src/content/4/en/part4b.md
@@ -727,6 +727,19 @@ test('a note can be deleted', async () => {
   expect(contents).not.toContain(noteToDelete.content)
 })
 ```
+In the test for 'a specific note can be viewed' we get an error because like the ID in 3b the date is stored as an Object in Mongoose without extra formatting but is compared to a JSON string.
+
+Let's add this to our models/note.js
+
+```js
+noteSchema.set('toJSON', {
+  transform: (document, returnedObject) => {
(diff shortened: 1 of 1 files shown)


src/content/4/en/part4b.md
```diff
@@ -723,6 +727,19 @@ test('a note can be deleted', async () => {
...
 ```
+In the test for 'a specific note can be viewed' we get an error because like the ID in 3b the date is stored as an Object in Mongoose without extra formatting but is compared to a JSON string.
+
+Let's add this to our models/note.js
+
+```js
+noteSchema.set('toJSON', {
+  transform: (document, returnedObject) => {
+  returnedObject.id = returnedObject._id.toString()
+  returnedObject.date = returnedObject.date.toString() // highlight-line
```
Kaltsoon, line 738: I would rely on the `new Date().toJSON()` implementation here, which is `new Date().toISOString()`

lelilia: looks less hacky than my attempt at fixing it with the knowledge of the course material :)

Kaltsoon:
The core issue is related to this piece of code:

```javascript
const notesAtStart = await helper.notesInDb()

const noteToView = notesAtStart[0]

const resultNote = await API
  .get(`/api/notes/${noteToView.id}`)
  .expect(200)
  .expect('Content-Type', /application\/json/)

expect(resultNote.body).toEqual(noteToView)
```

`resultNode.body` goes through JSON serialization and parsing whereas `noteToView` doesn't. This results in the date being a string in the `resultNode.body` and a `Date` object in the `noteToView`. The easiest way to fix this issue is to  change the code into:

```javascript
const processedNoteToView = JSON.parse(JSON.stringify(noteToView));

expect(resultNote.body).toEqual(processedNoteToView)
```

However, this requires an explanation.

Kaltsoon:
@lelilia are you able to check if the solution I provided actually works? Including the dates, you added to the note objects.

lelilia:
yes that does the trick \o/

Kaltsoon:
Awesome, thanks a lot! This has caused a lot of headache

PR #660, opened 2020-08-18
Update osa2b.md

p4nu:
Remove use of anti-pattern using var i as key and remove unneeded highlight as it's added to the code earlier


Kaltsoon:
Thanks!

PR #668, opened 2020-08-18
Update part3a.md

orvalho:
Fix typo. Two words as a verb (shut down), one word as a noun (shutdown). In this case, it is used as a verb - so two words.


Kaltsoon:
Thanks!

PR #682, opened 2020-08-21
Update part3c.md

orvalho:
If the json-parser middleware would be loaded after the logger middleware and the POST route handler - request.body would not be an empty object. It would be undefined.


Kaltsoon:
Thanks!

PR #693, opened 2020-08-25
Update osa3a.md

p4nu:
This PR makes the first example of creating a server a little easier to understand


Kaltsoon:
Good stuff, thanks!

PR #694, opened 2020-08-25
Modified description of exercise 6.19.

Translucent504:
Since the students have been using redux with the newer state hooks in part 6, the description of exercise 6.19 had the incorrect assumption that the redux store was being passed through props and that the students would have to install the 'react-redux' package. Very minimal changes have been made to update the exercise description so students would not be confused.

### DIFF
--- a/src/content/6/en/part6d.md
+++ b/src/content/6/en/part6d.md
@@ -606,11 +606,9 @@ More about this [here](https://www.simplethread.com/cant-replace-redux-with-hook
 
 #### 6.19 anecdotes and connect, step1
 
-The <i>redux store</i> is currently passed to all of the components through props.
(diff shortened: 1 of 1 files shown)


GreyGreyman:
Wanted to deal with the same problem before seeing your pr, great job!

The one thing should be addressed, imho, is that in the finnish version of 6.19 it seems that the task suggests to rewrite Notification component to use "connect", not AnecdoteList (not a finnish speaker, assumption made from google translate and messages from telegram chat of the course). 
Same thing with 6.20: english version asks to refactor Filter and AnecdoteForm components, and finnish one asks to refactor only AnecdoteForm component. I think we should take this opportunity to somehow unify those versions.

Kaltsoon:
Thanks, this will clarify things!

PR #702, opened 2020-08-27
Update part4d.md

orvalho:
Fix typo


orvalho:
@Kaltsoon 
Hey, maybe I should've added an explanation with my commit.  :blush:

So here it goes:
Sometimes words that begin with the letter H start with vowel sounds and sometimes they start with consonant sounds. 

For example, it is “a hotel” because “hotel” starts with a consonant sound (/**h**oʊˈtel/), but it is “an HTTP” because “HTTP” starts with a vowel sound (/ˌ**e**ɪtʃ.tiː.tiːˈpiː/). 

Besides HTTP term is used with the article 'an' on MDN - if that makes any difference.

PR #713, opened 2020-08-29
Modified Chinese Content According to the En Version(week35 @ 2020)

RichardStark:
Modified Chinese Content According to the En Version(week35 @ 2020)…Part 10 is under constructing(Finished 10a.md)


Kaltsoon:
Awesome work!

PR #716, opened 2020-08-29
Update part4d.md

orvalho:
Fix typo. Plural noun (applications) used with singular verb (stays). Change from 'applications stays' to 'application stays'.


Kaltsoon:
Thanks!

PR #719, opened 2020-08-30
Update part5b (all languages)

kimsappi:
Exercise 5.6 references the component NewNote, which doesn't exist in the material. Should be NoteForm instead.
However, it does appear that a NewNote component is found in at least part 6. For consistency's sake it might be better to change the NoteForm component to NewNote instead (that could also be more confusing).


Kaltsoon:
Thanks!

PR #729, opened 2020-09-03
Update part5a.md

orvalho:
Fix typo. Two words as a verb (log in), one word as a noun (login). In this case, it is used as a noun - so one word.


Kaltsoon:
Thanks!

PR #734, opened 2020-09-04
Update part10a.md

### LINKED ISSUE #733: Part 10a: expo command not found [windows]
After globally installing expo-cli with `npm install --global expo-cli`,
running `expo init rate-repository-app` leads to error:
> 'expo' is not recognized as an internal or external command,
> operable program or batch file.

For me this was the first time installing anything globally and the directory C:\Users\<USER>\AppData\Roaming\npm was not included in the Environmental variables. Adding the directory to path fixed the issue. Other operating system users _might_ experience a similar issue.

Consider adding a disclaimer regarding the topic.

tveijola:
Closes #733

### DIFF
--- a/src/content/10/en/part10a.md
+++ b/src/content/10/en/part10a.md
@@ -57,6 +57,8 @@ Next, we can initialize our project in a <i>rate-repository-app</i> directory by
 expo init rate-repository-app
 ```
 
+<b>NOTE</b> that to use the expo command, the npm root directory must be added to the system path. The path can be found by running the command ```npm root -g```
+
 After running this command Expo will ask you to choose a template for the project. Within the <i>Managed workflow</i> section, choose the <i>blank</i> option (the one with the description "a minimal app as clean as an empty canvas").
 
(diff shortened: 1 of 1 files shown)


Kaltsoon:
If the global installation doesn't work, maybe it would be better to instruct people to use `npx expo-cli init`. This however requires changes in the `package.json` scripts, e.g. `expo start` needs to be changed into `npx expo-cli start` and so on.

tveijola:
Thanks for the comment. The issue was not the installation itself, but the global root directory not being included in the system path.

Kaltsoon:
That's true. I was thinking what's the easiest way to execute the command if you can't run it globally. I'm afraid most of the people have no idea what "the npm root directory must be added to the system path"  means.

tveijola:
Oh yes, I didn't think of that. Is it possible to install expo locally in the project directory? The command could then be run only from the project directory. Would that cause more issues than solve?

sahilrajput03:
> Oh yes, I didn't think of that. Is it possible to install expo locally in the project directory? The command could then be run only from the project directory. Would that cause more issues than solve?

Installing expo globally is just better!

PR #737, opened 2020-09-04
Update part0b.md

orvalho:
"on the web" is predominantly used in the context of the World Wide Web. Therefore I suggest changing the preposition from "in" to "on". 

Changed from "Communicating **in** the web...." to  "Communicating **on** the web...."


Kaltsoon:
Thanks!

PR #738, opened 2020-09-05
Modified Chinese Content According to the En Version(week36 @ 2020), PART10 FINISHED!!!!

RichardStark:


Kaltsoon:
Awesome work, as always! 🥇

PR #739, opened 2020-09-05
Update part0b.md

orvalho:
Fix typo. 

The choice between a and an is determined by the initial sound, and not the initial letter, of the following word.

For example, it is “a hotel” because “hotel” starts with a consonant sound (/**h**oʊˈtel/), but it is “an HTML” because “HTML” starts with a vowel sound (/ˌ**e**ɪtʃ.tiː.emˈel/).


Kaltsoon:
Thanks!

PR #747, opened 2020-09-07
Fix part10c.md exercie number 10:14 for both finnish and chinese content in chapter 3, Part 10.

sahilrajput03:
Fix: en version: exercise no. 10:14(it was 10.12 earlier).


Kaltsoon:
Awesome, thanks!

PR #748, opened 2020-09-07
Update part0b.md

orvalho:
Based on the context I'm guessing that the material meant to say "**t**here" and not "**w**here"


Kaltsoon:
Thanks!

PR #750, opened 2020-09-07
Update part1a.md

orvalho:
The verb _want_ should be followed by  'to' + infinitive (to become) and not by a gerund (becoming).


Kaltsoon:
Thanks!

PR #754, opened 2020-09-08
Update part1b.md

orvalho:
I think that this sentence could be improved.

Changed
from:
"Today, the most popular way to do the transpiling is using Babel."
to:
"Today, the most popular way to do the transpiling is **by** using Babel."


Kaltsoon:
Thanks!

PR #758, opened 2020-09-09
Update part4c.md

twoinfiniteloops:
I might be wrong but there is a problem here.

The current code saves user to the db, but since we need to add notes to user, we only need to update user. The current version should give an error of "username appears" cause we "save" it again.

I propose update. As I could see, I updated the parts where it needs.

### DIFF
--- a/src/content/4/en/part4c.md
+++ b/src/content/4/en/part4c.md
@@ -462,7 +462,7 @@ notesRouter.post('/', async (request, response, next) => {
 
   const savedNote = await note.save()
   user.notes = user.notes.concat(savedNote._id) //highlight-line
-  await user.save()  //highlight-line
+  await User.findByIdAndUpdate(user._id, user, {new: true})  //highlight-line
   
   response.json(savedNote)
 })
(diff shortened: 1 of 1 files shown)


Kaltsoon:
Both `await user.save()` and `await User.findByIdAndUpdate(user._id, user, {new: true})` do pretty much the same, but in my opinion using the `user.save` method in this case is a much cleaner solution.

> The current version should give an error of "username appears" cause we "save" it again.

What do you mean by this? Is that error related to some validation, or what is it about?

twoinfiniteloops:
Yeah, it is about validation. When I first use "await user.save()", it gives an error of "username already appears". (Unique validation was an exercise in previous questions). I know when we use "save" in mongoose, it creates/updates an existing document. But, while doing exercise, I came across an error, so I want to notify you. But, it seems it is not a serious problem. Thank u so much, I am following the course with great passion, have a nice dayyy💯

Kaltsoon:
I'm glad to hear that you have enjoyed the course! I think in this case it would make more sense to fix the uniqueness validator so that the updates with the `save` method work. This probably just requires a check whether the user with a certain username is actually the one being updated, which shouldn't break the uniqueness validation.

abdallahakrab:
I tried to make .save() working with mongoose-unique-validator but didnt find direct way to do it and settled with await User.findByIdAndUpdate(user._id, user, {new: true})

PR #762, opened 2020-09-10
explain Finnish text in screenshot in part2a.md

johan456789:
I thought something's wrong when I see different text got print out. It'd be better to add a note that the screenshot is in Finnish.


sahilrajput03:
app tomini means "App works"
I don't know finnish too. I used google translate..:LOL:

PR #764, opened 2020-09-10
Update part1c.md

orvalho:
A function defining an _App_ component **doesn't have** a parameter **_props_** (line 254). 
However, when we later (line 309) take a closer look at the _App_ component - it now **has _props_**.
And if we read on (line 335) we see that the _App_ component **doesn't have _props_** again.

I suggest removing _props_ as it's not needed.


Kaltsoon:
Thanks!

PR #772, opened 2020-09-12
Update part1c.md

orvalho:
I believe that in this context it would be more accurate to use the word "function" and not "method".

A function is independent of an object.
A method is on an object.


Kaltsoon:
Thanks!

PR #773, opened 2020-09-12
Update part1c.md

orvalho:
I think that this sentence could be improved.

It isn't exactly clear to me what the word 'method' refers to. Consider using the word "component" instead.


Kaltsoon:
Thanks!

PR #775, opened 2020-09-12
Update part1d.md

orvalho:
Fix typo.

Someday = it talks about future events that will happen in an unknown time.
ex. _Someday_, I will become famous.

Some day = it talks about a 'specific' day or a 'single' day that is unknown.
ex. Let’s get lunch _some day_ next week.


Kaltsoon:
Thanks!

PR #779, opened 2020-09-13
Name of component must start with uppercase letter

sahilrajput03:
With name noteForm, the component won't even render at all.

### DIFF
--- a/src/content/5/en/part5b.md
+++ b/src/content/5/en/part5b.md
@@ -412,7 +412,7 @@ const App = () => {
   // ...
   const noteFormRef = useRef() // highlight-line
 
-  const noteForm = () => (
+  const NoteFormParent = () => (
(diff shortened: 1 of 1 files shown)


Kaltsoon:
The material doesn't explicitly demonstrate how to use the `noteForm` function. If the `noteForm` is called as a function (which I'm pretty sure is the point), like `noteForm()`, instead of as a component `<noteForm>` this is totally fine. I definitely wouldn't encourage people to define components inside a component's render function.

sahilrajput03:
I though it different way, actually I totally forgot that I could render components via function calling. 

I am fine with the course's approach now. :)

But the way I analysed at that time was more like shown in below screenshot(link to sandbox below), and bcoz of that error - 

src: https://codesandbox.io/s/functioncall-rendering-in-reactfso-ciu4n

image
Thanks, I'm happy and satified with your approach, 
Thanks for replying to this thread.

PR #792, opened 2020-09-16
Hour range to complete parts added

surfmuggle:
I suggest that a phrase like this "In 2020 on average the parts took from 8.8 hours (part0) up to 53.2 hours (part9). " is added.


Kaltsoon:
This is an ok idea, but a bit tedious to maintain from year to year

PR #795, opened 2020-09-17
Update part3c.md

johan456789:
This is the first time `findById` is used, so there should be a reference to Model.findById().


Kaltsoon:
Cool, this is a good improvement. Thanks!

PR #799, opened 2020-09-18
Part 10 changes

yusifsalam:
First commit fixes line highlighting.
Second commit fixes incorrect path for apolloClient.js within App.js. App.js is in the root folder, utils folder is a subfolder of src, so ./utils relative to App does not exist. 
Third commit fixes some typos.


Kaltsoon:
Thanks!

PR #802, opened 2020-09-20
Update part3c.md

ruikawahara:
Current instruction on how to use Visual Studio Code debugger is possibly out of date. For example, my VS Code did not have "Debug" option, but it had "Run" instead. Furthermore, directly running the debugger caused chrome tab to open on wrong port. Even after fixing the port via launch.json, debug panel would not let me observe the changes.

Upon tweaking  my VS Code by adding  "npm start" configuration to launch.json, I was able to use my debugger properly. 

Please note that I am proposing above changes based on my personal experience. I've also attached a screenshot of what my working debugger looks like for reference.

pull_req_part3C_debug


Kaltsoon:
Thanks for pointing this out!

PR #804, opened 2020-09-20
Update osa4b.md

p4nu:
This PR updates code to use currently used mechanism to check if the new note's importance is defined when adding a new note in notesRouter.post function.


Kaltsoon:
Thanks!

PR #805, opened 2020-09-20
fix(part 10): prevent application crash

zephimir:
The empty string is causin the application to crash:
> Error: Text strings must be rendered within a <Text> component.

image

Removing it fixes the issue


Kaltsoon:
Thanks! I have no idea how that `{' '}` ended up there.

zephimir:
Happened to me a few times, it's Prettier doing weird stuff 😐

PR #810, opened 2020-09-21
docs(part10a.md): fix react native debugger port

jeremy-ebinum:
According to the expo docs https://docs.expo.io/workflow/debugging/#developer-menu the React Native Debugger should run on port 19001 not 9001


Kaltsoon:
Good catch, thanks!

PR #812, opened 2020-09-22
Update osa4b.md

p4nu:
This PR fixes forgotten --save-dev flag when installing cross-env dependency. According to their official docs it should be installed with that flag.

https://www.npmjs.com/package/cross-env#installation


Kaltsoon:
Thanks!

PR #813, opened 2020-09-22
Update osa4b.md

p4nu:
This PR simplifies the beforeEach function to use insertMany function in the refactoring of tests.


Kaltsoon:
Definitely an improvement. Thanks!

Kaltsoon:
Actually, there's a reference to this later on:

> Toimiva ratkaisu tilanteessa on odottaa asynkronisten talletusoperaatioiden valmistumista beforeEach-funktiossa, esim. metodin Promise.all avulla

There might be a pedagogical reason to do this the hard way.

p4nu:
I noticed it was defined like this earlier in this part but if it's defined in later sections let's leave it like this 👍

PR #818, opened 2020-09-24
Update part7b.md

twoinfiniteloops:
I think it is a bug ?


Kaltsoon:
`./hooks` import is a shorthand for `./hooks/index` (both resolve the same file), so there's no problem here.

twoinfiniteloops:
thx

PR #820, opened 2020-09-24
Update part4d.md

alumezi:
I think all tests should pass before continuing to the next part.


Kaltsoon:
We aren't adding new exercises at the moment. We will consider this in the future.

Issue #449, opened 2020-05-28
Part 2c: npm --save deprecated


danieljimenez1337:
The use of the --save flag is deprecated for npm version 5 and up. Npm now saves dependencies by default. This is done multiple times throughout the course.
 
![2020-05-27_19-45](https://user-images.githubusercontent.com/40440627/83092759-d8021900-a052-11ea-9357-da307b9c5a81.png)

The --save-dev flag is correct usage.

waciumawanjohi:
To add:
The course states that code samples use Node 10.18.0 [link](https://fullstackopen.com/en/part1/java_script).
Node 10.18.0 is packaged with npm 6.13.4 [link](https://nodejs.org/en/blog/release/v10.18.0/)
npm documentation of the install command does not include the --save flag. It notes that "npm install saves any specified packages into dependencies by default." [link](https://docs.npmjs.com/cli-commands/install.html)

Kaltsoon:
Thanks for pointing this out! I've removed the use of the `--save` flag from material. If you still find incorrect usage in some parts of the material, please open a pull request.

PR #821, opened 2020-09-25
Update osa4c.md

p4nu:
This PR adds bcrypt to be declared as dependency in package.json


Kaltsoon:
The `--save` flag isn't necessary: https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/issues/449. Did you face issues without the flag?

p4nu:
I was not aware you don't need the flag anymore 😅

PR #827, opened 2020-09-26
exprees is a abstraction of node's http module.

sahilrajput03:
Just a recommendation to add this to course, coz I never know express was built on top http module itself even after reading all that useful information in the course.  Material uses the word ``than`` which kind of makes me think it says express is different from node's http, but express is just abstraction of http' node moudle.
Thanks!


Kaltsoon:
Thanks!

PR #837, opened 2020-09-28
Update osa4d.md

p4nu:
There is no step3 defined in earlier parts so this PR updates the step numbers in exercises


Kaltsoon:
Thanks!

PR #850, opened 2020-10-02
Part 2b, setting state

cdrbbt:
I've commented out a part that seems to be a bit confusing. I feel like it would be best to explain why is it fine to modify 2 states in a row when the first setState call would in theory immediately rerender the component


Kaltsoon:
Commenting the code would break the application, why is this an improvement?

PR #851, opened 2020-10-02
Update part0b.md

Brentably:
the address changed so what the description says doesn't match the photo, so it might be worth clearing up that discrepancy


Kaltsoon:
Thanks!

PR #856, opened 2020-10-05
part2c highlight change

cdrbbt:
This line has also changed since notes are now not passed as props to the App component


Kaltsoon:
Thanks!

PR #860, opened 2020-10-07
Part3a inconsistent variable names

cdrbbt:
This piece of code is referenced 2 more times later, however there the variables are request and response instead of req and res


Kaltsoon:
Thanks!

PR #861, opened 2020-10-07
part3 inconsistent variable names2

cdrbbt:
When the code is first declared the variables are named req and res, but when the code is referenced again they are refereed to as request and response


Kaltsoon:
Thanks!

PR #867, opened 2020-10-10
Update part2c.md

lidaniel7:
Added additional clarification for which console to navigate to. I was stuck for some time on this lesson because I thought the console logs in the image corresponded to localhost:3001 where the server is at as opposed to localhost:3000 where the React application is.


Kaltsoon:
Thanks!

PR #870, opened 2020-10-10
Update part1d.md

orvalho:
When specifying attributes with JSX:
* use quotes for string values
* use curly braces for expressions

But not both in the same attribute.

From React docs:
[Specifying Attributes with JSX](https://reactjs.org/docs/introducing-jsx.html#specifying-attributes-with-jsx)


orvalho:
_label:hacktoberfest-accepted_ appreciated

Kaltsoon:
Thanks!

PR #871, opened 2020-10-10
Update part1d.md

orvalho:
When our event handler is defined as a _function call_, the learning material says: 
"The message gets printed to the console once but nothing happens when we click the button a **second time**."

In fact, nothing happens when the button gets clicked the first time also. The message gets printed to the console only once when the component is rendered.

### DIFF
--- a/src/content/1/en/part1d.md
+++ b/src/content/1/en/part1d.md
@@ -587,7 +587,7 @@ What about the following:
 </button>
 ```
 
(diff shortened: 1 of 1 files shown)


orvalho:
_label:hacktoberfest-accepted_ appreciated

Kaltsoon:
Thanks, this will clarify things!

PR #872, opened 2020-10-10
Update part1d.md

orvalho:
In my opinion, there is no need to use nested divs in this case. It adds no benefit. 
Eliminating unnecessary divs makes code look lean, clear and free of fluff.


Kaltsoon:
Definitely true, thanks!

PR #873, opened 2020-10-10
Update part1d.md

orvalho:
A function defining an App component has a parameter props. 
However, no props are passed to the App component.
We don't need to add the props parameter in the function definition unless we want to use it.
Therefore, I suggest removing props as it's not needed.


Kaltsoon:
Thanks!

PR #877, opened 2020-10-11
Update osa8d.md

anttipessa:
Fix to a broken link. If accepted please add tag `label:hacktoberfest-accepted`


Kaltsoon:
This source is perhaps better: https://www.apollographql.com/docs/react/data/queries/#configuring-fetch-logic

PR #879, opened 2020-10-11
Update part2a.md

orvalho:
I think that this sentence could be improved by not using the word “topic” two times in one sentence.


Kaltsoon:
Thanks!

PR #888, opened 2020-10-12
Update part6c.md

rhysmitchell:
Added --save flag to include axios as a dependency in the package.json.


Kaltsoon:
`--save` flag is not necessary

PR #890, opened 2020-10-12
Update osa1d.md

TarmoJussila:
Fixed word 'tila' in Finnish version of part1d (osa1d). Word 'tilaa' was used which is not right.


TarmoJussila:
Also fixed typo 'muuteaan' to 'muutetaan'.

Kaltsoon:
Thanks!

PR #891, opened 2020-10-12
Update osa7d.md

jpparkkari:
webpack.config.js has 'Rule.query', which is deprecated already in webpack 4 and not working in webpack 5. 'Rule.options' works instead


Kaltsoon:
Thanks this is fixed

PR #892, opened 2020-10-12
Update part9b.md

prrdee:
It should be height and weight instead of weight and height to get the result shown in instruction.


Kaltsoon:
This makes sense. Thanks!

PR #893, opened 2020-10-12
Update part7d.md

jpparkkari:
Rule.query is not supported in webpack 5 and deprecated in webpack 4. Replaced that with Rule.options as said in https://webpack.js.org/configuration/module/#ruleoptions--rulequery


Kaltsoon:
Thanks!

PR #894, opened 2020-10-13
Update osa7d.md Rule.loaders -> Rule.use

jpparkkari:
Rule.loaders is deprecated and does not work in webpack5. Changed it to Rule.use as instructed in https://webpack.js.org/configuration/module/#ruleloaders
This will also work, if it would be educationally better:
use: [
          {loader: 'style-loader'}, 
          {loader: 'css-loader'}
        ],


Kaltsoon:
Thanks!

PR #897, opened 2020-10-13
Update osa7d.md webpack-cli version

jpparkkari:
Fixed the webpack-cli version to 3.3.12, as webpack-cli is now by default at version 4.0 and that does not work with webpack-dev-server. Easiest way around seems to be to install the latest v3 package. Otherwise webpack-dev-server needs to be changed to webpack serve. And that will need more changes to material.


Kaltsoon:
After doing some testing myself, I think changing the `start` script into `webpack serve --mode=development` (which you already mentioned) is the best idea and requires the least changes. I'll make this small change to each translation. Thanks for the help!

PR #900, opened 2020-10-13
Update part2a.md

orvalho:
I believe that note (array item) and not Note (component) should be used in this context.


Kaltsoon:
Thanks!

PR #901, opened 2020-10-13
Update part2b.md

orvalho:
Fix typo.
A hyphen is used to create compound words. 
The two words joined by the hyphen act as an adjective (user-submitted) that describes a noun (input) in the sentence.


Kaltsoon:
Thanks!

PR #907, opened 2020-10-14
Update part2c.md

orvalho:
Fix typo.

noun - practi**c**e
verb - practi**s**e/practi**c**e

In this case, it is used as a noun - therefore practi**c**e


Kaltsoon:
Thanks!

PR #910, opened 2020-10-14
Update part2c.md

orvalho:
Fix typo.
A hyphen is used to create compound words.
The two words joined by the hyphen act as an adjective (machine-readable) that describes a noun (format) in the sentence.


Kaltsoon:
Thanks!

PR #912, opened 2020-10-15
Update part4b.md

ruikawahara:
```express-async-errors``` library should be imported to ```app.js``` rather than ```src/app.js```.

This is under "Eliminating the try-catch" section. Modified line 801 of ```part4b.md```.

I have also double checked and noticed that the branch containing completed result of above section does not have ```src``` directory.
Source: https://github.com/fullstack-hy2020/part3-notes-backend/tree/part4-5


Kaltsoon:
Thanks!

PR #913, opened 2020-10-15
Update part5a.md

YousifAbozid:
this is a little confusing for me, because we are sending a request from the frontend to the backend, so I guess it should be a request not response, I know that this won't harm the code that much and everything will work perfectly, but I prefer to correct it to prevent interference, thanks :)


Kaltsoon:
We send requests, but the server responds with a [response](https://github.com/axios/axios#response-schema)

YousifAbozid:
I guess I understand now, so there is no need for my changes, Thanks Kalle :) you can close the pull request now

PR #916, opened 2020-10-16
Update part3c.md

lidaniel7:
The URI address should contain `${password}` instead of `<PASSWORD>` for the code to run correctly.


Kaltsoon:
`<PASSWORD>` is just a placeholder. `${password}` would require the use of a template string and define a `password` variable. This would potentially confuse a lot of people

PR #921, opened 2020-10-16
Content --> Part

jayeshmann:
Acc. to me, this exercise want students to add the additional case after adding interface, union and data. Because, without it, it would lead to error. 
Switch case will be added to Part component instead of Content component. Please consult the author of this exercise, if possible, before merging.


mluukkai:
no more relevant since the exercise has changed a bit

PR #930, opened 2020-10-20
Update part3d.md

KaungKhantThar7777:
Mongoose now support unique key in schema.


Kaltsoon:
Can you provide a source for this?

KaungKhantThar7777:
Screenshot_2020-10-22-22-13-51-375_com android chrome

Screenshot_2020-10-22-22-13-04-436_com android chrome

Although unique is not a validator,we can use it in schema.It is thrown an exception while unique is violated? Please forgive me if I was wrong.I am just learning now.

PR #937, opened 2020-10-24
where user -> where the user

t-miller:
Change `where user` to `where the user` to match the rest of the questions.


Kaltsoon:
Thanks!

PR #941, opened 2020-10-25
Update create-react-app filenames

t-miller:
It appears as though `serviceWorker.js` has now become `reportWebVitals.js`


Kaltsoon:
That seems to be true. Thanks!

PR #952, opened 2020-10-30
npm install mixed with yarn syntax

Matheos96:
Changed `npm add` to `npm install`. Faulty syntax likely sprung from yarn syntax (`yarn add` being the yarn equivalent of `npm install`)


Kaltsoon:
Thanks!

PR #958, opened 2020-11-03
Update part3c.md

coldPen:
In practice, use of the `toJSON` method doesn't seem mandatory to get cleaned JSON formatted data with a custom `id` field, but the course does mention it and the examples don't include it, I don't understand why.

Please either rewrite the relevant instructions or include the method in the examples. In case of a complete rewrite, it would be interesting to add an explanation as to why mongoose seems to implicitly use our modified `toJSON` method.

This problem causes discrepancies between code examples in part 3c and part 3d, where we can see `toJSON` used in snippets written in previous parts of the course.

### DIFF
--- a/src/content/3/en/part3c.md
+++ b/src/content/3/en/part3c.md
@@ -427,7 +427,7 @@ Let's respond to the HTTP request with a list of objects formatted with the _toJ
 ```js
 app.get('/api/notes', (request, response) => {
   Note.find({}).then(notes => {
-    response.json(notes)
+    response.json(notes.map(note => note.toJSON()))
   })
 })
 ```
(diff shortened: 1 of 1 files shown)


Kaltsoon:
Mongoose doesn't implicitly call the `toJSON` method, it is done by [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) (called by express). It is mentioned here (the first time `toJSON` method is implemented): https://fullstackopen.com/en/part3/saving_data_to_mongo_db#backend-connected-to-a-database

> Now the notes variable is assigned to an array of objects returned by Mongo. When the response is sent in the JSON format, the toJSON method of each object in the array is called automatically by the JSON.stringify method.

If there's room for improvement, please let us know.

Kaltsoon:
I agree that the [Promise chaining](https://fullstackopen.com/en/part3/validation_and_es_lint#promise-chaining) example (which you are probably referring to) is a bit confusing because it calls the `toJSON` method. We should come up with another promise chaining example.

PR #970, opened 2020-11-13
Add usage of `immerjs` for `store updation` in apollo cache, as it makes code more readable. 😎

sahilrajput03:
Add introduction to immerjs if you feel necessary.


sahilrajput03:
https://docs.gitlab.com/ee/development/fe_guide/graphql.html#immutability-and-cache-updates

I referenced the usage from here.

Thanks.

PR #974, opened 2020-11-18
fix faq broken links

dgrishajev:
Mainly due to this [commit](https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/commit/b8ac9d97c8b9907c5f1e06309434b274ddd24b15) links from FAQ sections on expanding on a previous course are now leading to 404 (or anchor is not working). This commit fixes the URLs and also syncs the typo changes between ZH and EN parts. Also, my IDE seems to have removed some extra spaces.


Kaltsoon:
Thanks!

PR #986, opened 2020-11-27
Update part5c.md: Make getByText as a method of component

georgemihail:
I think it's need to mention that getByText is a method of component.
So, let's add 'component' before.


Kaltsoon:
Thanks!

georgemihail:
My pleasure!

PR #995, opened 2020-12-03
Revive React Native documentation links

olav35:
React Native documentation prior to version 0.60 has been moved from https://reactnative.dev/docs to https://archive.reactnative.dev/docs. This PR revives two dead links by making them point to https://archive.reactnative.dev/docs/*, instead of the now outdated https://reactnative.dev/docs.


Kaltsoon:
Thanks, I think we should use the docs of the latest version.

PR #1006, opened 2020-12-08
Update part3c.md

hpasweden:
I think an underscore went missing


hpasweden:
But later saw it was rewriting the key from _id to id :)

PR #1022, opened 2020-12-10
Update part11d.md

YousifAbozid:
fixed some typos also the hash was wrote wrong as v1.33.0 in this commit
https://github.com/anothrNick/github-tag-action/commit/eca2b69f9e2c24be7decccd0f15fdb1ea5906598
there is no number 9 in the beginning of the hash as you see above


mluukkai:
👍

PR #1023, opened 2020-12-10
Update part11a.md

YousifAbozid:
typo


mluukkai:
this is actually a intended "typo"

PR #1028, opened 2020-12-11
Typofix.

samilaine:


sarthakgaur (changes requested):
I think you probably meant, `we do not have much time for this now either`.

PR #1071, opened 2020-12-29
Fix usage of `toJSON()` in Part 3c

abmerille:
Updated to match associated text and code in github for these sections.


Kaltsoon:
Calling `toJSON` method isn't required, it is done automatically by [JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)-method

PR #1083, opened 2021-01-10
Update part4c.md

Jordan-Rob:
on line 259  we reference  ``` body.password  ```  yet in the user schema there's an attribute of passwordHash instead. For better clarity, I see it much better to instead name the property passwordHash defined in the user schema to password since in the tests created later on for example on line 315 and 367  a new user is created with a property of password and not passwordHash.
on line 264   we directly place the passwordHash created in a new User object without assigning it to the appropriate user property.


kayraberktuncer:
passwordHash has been deleted, but passwordHash is used in many places. Shouldn't they change?

PR #1093, opened 2021-01-14
The Tests are note working

96berry96:
From here i can't continue the course, because the tests are note working,
and without the tests there is no way to continue the course.

I really like the course, you guys are greate teachers and i hope you guys can fix this 
Thaks


mluukkai:
well, there must be something wrong in your setup. it has been verified many times that the code in material works

PR #1101, opened 2021-01-19
Browsers actually know how to load modules

fcole90:
As proof, I developed an Aalto project that only uses ES6 modules, and I divided the code into multiple files, of which I only load the entry point: https://fcole90.github.io/interactive_bayesian_optimization/study.html

You can see at the bottom
```html
<script type="module">
  import {ui} from "./static/libs/ui.js";
  let app = new ui.Application();
  app.exec();
</script>
```

JS modules: https://github.com/fcole90/interactive_bayesian_optimization/tree/master/interactive_bayesian_optimisation/static/libs

### DIFF
--- a/src/content/7/en/part7d.md
+++ b/src/content/7/en/part7d.md
@@ -16,7 +16,7 @@ We can not rely on the black magic of create-react-app forever and it's time for
 ### Bundling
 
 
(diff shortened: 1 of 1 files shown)


Kaltsoon:
Thanks!

PR #1126, opened 2021-02-03
Change link to description as the repo has no info

fcole90:
G. Canti's `io-ts` repo has no info on what it is and how it works. The new link provides such information.
https://gcanti.github.io/io-ts/


Kaltsoon:
Thanks!

PR #1145, opened 2021-02-17
Fixed a missing return for the Button component in the Conditional rendering section of Part1-d

Shigeyuki-fukuda:
## Summary

In the Conditional rendering section of Part1-d, the return of the Button component is missing, causing the following error.

スクリーンショット 2021-02-17 22 38 31

Fixed the documentation by adding the return statement for the Button component so that the error does not occur. 
https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/commit/71544362ed5c3656aaa1a6280cc21b81c0d79def

### DIFF
--- a/src/content/1/en/part1d.md
+++ b/src/content/1/en/part1d.md
@@ -319,9 +319,11 @@ const History = (props) => {
 
 // highlight-start
 const Button = ({ onClick, text }) => (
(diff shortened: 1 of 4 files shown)


Shigeyuki-fukuda:
I made a mistake with {} and (). I've solved it and will close it. 🙏

PR #1150, opened 2021-02-21
Add support for using `cra` as an alias.

sahilrajput03:
Using cra alias would allow users to create cra apps really handy instead of writing whole lot of text while creating simple react apps.


Kaltsoon:
Too much hassle, the command is only needed a few times

PR #1157, opened 2021-02-28
Add note on RNCAndroidDialogPicker

fcole90:
Explains how to get it to work when it doesn't.
https://github.com/react-native-picker/picker/issues/136#issuecomment-693013536


Kaltsoon:
The new version of the material suggests using [@react-native-picker/picker](https://docs.expo.io/versions/latest/sdk/picker/)

fcole90:
You're right, and in my case it didn't work, possibly beacuse I could not install it using expo install, to keep all the other libraries in a compatible manner with rect native. Installing with yarn or npm worked, but then importing the component made the app to crash: [see here](https://github.com/react-native-picker/picker/issues/136#issuecomment-688505408)

On my device, even loading their snack made expo app to crash.

Using the legacy `"@react-native-community/picker": "1.6.0"` is the only working solution I could find after hours of research :sweat_smile: That's indeed a workaround, not a solution, but I think in this exercises it can alleviate some headaches.

PR #1163, opened 2021-03-08
Part11 exercise numbering skips 11.13

PTParta:
Part11 material has no 11.13 exercise.


mluukkai:
will be fixed in 2022 edition

PR #1170, opened 2021-03-11
Fixed link

didzis1:
Link was not clickable on 'exercise submission system'. This was caused by an extra parenthesis before the link itself.


Kaltsoon:
Thanks!

PR #1172, opened 2021-03-11
Update part7a.md

katamatata:
I've just finished exercises from part 7a. 
id and match.params.id are strings and unless I delete Number() function I get undefined here:

const note = notes.find(n => n.id === Number(id))

and here:

const note = match 
    ? notes.find(note => note.id === Number(match.params.id))
    : null

### DIFF
--- a/src/content/7/en/part7a.md
+++ b/src/content/7/en/part7a.md
@@ -292,7 +292,7 @@ import {
 
 const Note = ({ notes }) => {
   const id = useParams().id // highlight-line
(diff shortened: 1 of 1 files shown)


Kaltsoon:
Note ids seem to be numbers: https://fullstackopen.com/en/part6/flux_architecture_and_redux#redux-notes. But anecdote ids seem to be strings: https://github.com/FullStack-HY/routed-anecdotes/blob/main/src/App.js#L87. So the `Number(id)` shouldn't be used in the latter case.

PR #1186, opened 2021-03-17
Change in one of the headings

warty-warthog:
Updated " Tilan vieminen alikomponenttiin " to " Vie tila alakomponenttiin ". This is because " Tilan vieminen alikomponenttiin " translates as "Exporting space to subcomponents" in English while " Vie tila alakomponenttiin " translates as "exporting state to subcomponents". I feel like it should be state, not space


mluukkai:
It does not work like this. Finnish is the original language.

PR #1188, opened 2021-03-17
Removed an extra 'the' from a paragraph on the Old React section

Jetlum:
A small step for me a giant leap for future users


Kaltsoon:
Thanks!

PR #1189, opened 2021-03-17
Sentence rewording

Jetlum:
Reworded the sentence and removed 'your', with 'the' it made more sense and the word 'you' should not be used in these kind of texts, makes it seem like the text is directed to the person reading the sentence and not a general knowledge/information.


Kaltsoon:
Thanks!

PR #1193, opened 2021-03-19
Changing React Native Debugger port to 19000

akiratoma:
I couldn't make it work with port 19001. Reading the documentation: https://docs.expo.io/workflow/debugging/#startup, it seems like for newer versions, the port should be set to 19000. I downloaded the Expo Go app version 2.18.7 for Android.


Kaltsoon:
Good catch, thanks!

PR #1206, opened 2021-03-20
Update part3b.md

Birjot-Bala:
Fixed typos


exitthehouse (changes requested):
@Birjot-Bala , I don't have permission to commit to your branch, but I suggest you flip `now` and `resides` so the sentence reads: "The node/express-backend now resides in the Heroku server.

Birjot-Bala:
@exitthehouse Thanks, I have made the change you suggested.

PR #1207, opened 2021-03-21
Removing @react-native-community/async-storage

akiratoma:
I assume this is just an outdated import. The @react-native-community/async-storage package has been depracated: https://www.npmjs.com/package/@react-native-community/async-storage


Kaltsoon:
Yes, thanks!

PR #1209, opened 2021-03-22
Update part4b.md

khanhchung1997:
HTTP POST should return status 201 if it succeeds

### DIFF
--- a/src/content/4/en/part4b.md
+++ b/src/content/4/en/part4b.md
@@ -476,7 +476,7 @@ test('a valid note can be added', async () => {
   await api
     .post('/api/notes')
     .send(newNote)
-    .expect(200)
+    .expect(201)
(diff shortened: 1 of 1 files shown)


Ashu-r:
Hi, @khanhchung1997 as from [this](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_success) wikipedia page,

**200 OK**
    _Standard response for successful HTTP requests. The actual response will depend on the request method used. In a GET request, the response will contain an entity corresponding to the requested resource. In a POST request, the response will contain an entity describing or containing the result of the action._
**201 Created**
    _The request has been fulfilled, resulting in the creation of a new resource._

The 201 is for created and we have not created anything here.

mluukkai:
fixed in another commit

PR #1218, opened 2021-03-28
Update part1b.md

tiln7:
Screenshot at Mar 28 11-03-26


mluukkai:
well, this change would in runtime error. it was like it is meat to be

PR #1236, opened 2021-04-05
Update part4c.md

kayraberktuncer:
I removed this line, because there is no link.


mluukkai:
will be fixed in 2022 edition

PR #1237, opened 2021-04-06
fix log position

sahilrajput03:


sahilrajput03:
This PR fixes two things:
1. The log for `console.log('Person.find_v1')` is actually placed wrong in current material i.e., 

image

image

2. Make the log statements explicit that in either case i.e., if you **provide the phone number** or **not,** absurd queries are **supposed to happen in each case** as that is not explained in material.

Thanks.

sahilrajput03:
@mluukkai , Can you review the PR as its been here from a long time too and sorry i forgot to add details of what i requested in the PR at the time of creating the PR.

sahilrajput03:
Great, @mluukkai !!
Thanks. :)

PR #1265, opened 2021-04-22
Fix critical error in new exercise.

sahilrajput03:
Read the most voted comment of [this answer](https://stackoverflow.com/a/32572159/10012446) to know about the issue.


sahilrajput03:
Added

- [x] Differenet way to get _chat id_ for the group.
- [x] Add info about sending message to userinfobot from the mobile using the link in the material.
- [x] Simplify structure for reading the process.

PR #1267, opened 2021-04-23
Update part4a.md

Ishmam156:
Since express can take care of the HTTP requests, removing the node `http` in the index.js file as it might create confusion since this way was not used in part3.


mluukkai:
this change would require lots of changes to the example code in github so will not do this for this years edition

PR #1281, opened 2021-05-02
Corrected typo

NWilde4:
'secret' instead of 'secred'


mluukkai:
well, it was intentional typo but why not change

PR #1323, opened 2021-05-22
Updated deprecated link

Ashu-r:
When the link for type any (https://www.typescriptlang.org/docs/handbook/basic-types.html#any) is visited\,
The message is shown on the page as:
```
This page has been deprecated

This handbook page has been replaced, go to the new page
```
So the link has been updated with new handbook (2) documentation.


Ashu-r:
Further links in part 9b needs to be updated.  #1325

PR #1332, opened 2021-05-26
Adding Bind Method in Class Component

JustinKim3420:
A bind method is needed within the class component if the method is used in JSX. this.handleClick() should be undefined within the render method.


Kaltsoon:
`bind` call is not needed while using the arrow function while defining the `handleClick` method.

PR #1343, opened 2021-06-03
Update part2c.md

zacharydub:
inserting 'npx' to the script in package.json file before 'json-server...'


mluukkai:
you do not need it there if/when json-server is among dev-dependencies

PR #1350, opened 2021-06-09
Remove the "&& git push" warning

edgarasben:
I did not have a remote repo in GitHub, so the "deploy:full" script did not work.


mluukkai:
change in another commit

PR #1376, opened 2021-06-29
Addition of relative path for index.js

Reyno851:
Proposing addition of relative path for index.js as it is not usually placed in nodemon/.bin folder but in root directory


mluukkai:
this is not needed.

PR #1397, opened 2021-07-13
Fix grammar part2 Forms

09try:
Change persons to people


mluukkai:
will change this in next years edition. just changing this file is not enough so the change is not trivial

PR #1413, opened 2021-07-27
Update part8e.md

JoseDRamirezM:
In the last section of part 8 GraphQL subscriptions, while reading the material and running the code, I noticed that the implementation showed as example no longer works. So i decided to go through Apollo documentation (wich isn't quite clear) and found an example of a working implementation of the server [here](https://www.apollographql.com/docs/apollo-server/data/subscriptions/#enabling-subscriptions). I make this pull request in order to update the material so that other learners don't get stuck.


mluukkai:
fixed by a rewrite of the chapter

PR #1416, opened 2021-07-29
Fix bash commands in osa2c.md

qmisan:
In part 0 bash commands were said to start with _$ _mark. In this part they did not start with that. Also bash commands had here js coloring, fixed that also.


Kaltsoon:
Most bash commands don't have the `$` prefix and I think adding it would just cause copy-paste errors. Adding `bash` to the code highlighting on the other hand is a good idea.

PR #1418, opened 2021-07-29
Update part 8e.md

wade-dao:
Adding changes for the server: 
- Switch from apollo-server to apollo-server-express
- Add & import all other required dependencies: express, subscriptions-transport-ws, @graphql-tools/schema, graphql
- New logic for creating SubscriptionServer - starting from Apollo Server 3


mluukkai:
this is a too big change and not in sync with Finish version

PR #1436, opened 2021-08-09
Update part0b.md

geonidas:


mluukkai:
this info needs to be deduced by each student

PR #1439, opened 2021-08-10
corrected the class that was misspelled

Lenugo:


Kaltsoon:
Thanks! Could you get rid of the code snippet changes (e.g. semicolons), I don't think they are necessary

Lenugo:
> Thanks! Could you get rid of the code snippet changes (e.g. semicolons), I don't think they are necessary

Done!

PR #1459, opened 2021-08-23
Small fix to REST client in part3.md

PedroYanezE:
I had an issue when using REST client as indicated in the course: the requests were being sent to 127.0.0.1 instead of localhost:3001. Searching the web I found a closed issue on the plugin's Github page, where the following response solved the problem for me: https://github.com/Huachao/vscode-restclient/issues/122#issuecomment-673163671.
I added the fix in the shortest way I could! Hope this helps somebody


PedroPapelillo:
I apologize because I just realized that on Part 3's 'b' section the command for app.listen(...) changes for heroku, I had to change it aggain from what I stated on this pull request. Anyways, it could be worth something to mention that you have to pass 'localhost' as a parameter for app.listen(...) to use REST Client **locally**. Sorry for any inconvenience!

PR #1477, opened 2021-08-31
Remove mongoose.connect() options parameter for

SupremeChow:
Going through the tutorial, Node give errors regarding options parameters of mongoose.connect(url, { ... }). Going off https://mongoosejs.com/docs/migrating_to_6.html#no-more-deprecation-warning-options for Mongoose 6.0, the options are true by default and should be removed.


Kaltsoon:
Thanks we have fixed this!

PR #1478, opened 2021-09-02
Remove usefindandmodify and usecreateindex from mongoose.connect()…

Junah-Park:
…function call parameters

UnhandledPromiseRejectionWarning: MongoParseError: options usefindandmodify, usecreateindex are not supported


Kaltsoon:
Thanks we have fixed this!

PR #1492, opened 2021-09-11
fixed typo in code : _id property must be assigned

Vinayaka2k:
In lines 417 and 469, _id property must be assigned while converting the _id object to String.


mluukkai:
this is not a typo

Vinayaka2k:
Even if its not a typo, returnedObject.id also would be returned from the backend and there would be an additional id property along with content and important properties, which I think is undesirable in this case

PR #1502, opened 2021-09-23
Update and rename src/content/0/es/part0b.md to src/contenido/0/es/pa…

WilmarDeML:
…rt0b.md


mluukkai:
file names can not be changed, would break things

PR #1509, opened 2021-09-28
Added warning regarding node bug 10638

gkowalc:
When using default node version installed from ubuntu 20.04 repo (10.x version)   running comand node mongo.js <password> following error will be displayed:
const utf8Encoder = new TextEncoder();
ReferenceError: TextEncoder is not defined

I was able to find corresponding bug report:
https://github.com/Automattic/mongoose/issues/10638

Also I was able to run the  same code without any problems through heroku console.


Kaltsoon:
The material doesn't support Node version 10.x

PR #1513, opened 2021-09-30
Update part12a.md

NWilde4:
`docker-fs: command not found` on WSL. 
`docker image ls`  works fine and shows the correct output.


Kaltsoon:
Thanks!

PR #1515, opened 2021-10-03
Add additional task for part6d

Laute:
Instructions in part6d ask you to modify the `Notification`, `Filter`, and `AnecdoteForm` components to use the connect function rather than hooks, but this leaves `AnecdoteList` alone unmodified at the end of the part. I think it's a worthwhile exercise to modify `AnecdoteList` as well since it's the only component in this project that maps both states and dispatches to props when using connect.

In my commit I've added this extra task to exercise 6.20, but it could alternatively be added as a separate exercise (e.g. as exercise 6.21, then `anecdotes, the grand finale` gets pushed to exercise 6.22).


mluukkai:
part 6d has changed in 2022 edition

PR #1516, opened 2021-10-04
Fix typo

rajshekhar26:
Replace '--only-production' with '--only=production'

### DIFF
--- a/src/content/12/en/part12b.md
+++ b/src/content/12/en/part12b.md
@@ -89,7 +89,8 @@ Containerizing that should be relatively easy based on the previous example.
 - Use node as base
 - Set working directory so we don't interfere with the contents of the base image
 - Copy ALL of the files in this directory to the image
-- Start with DEBUG=playground:* npm start
+- Start with DEBUG=playground:* npm startaterial
+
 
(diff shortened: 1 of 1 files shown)


src/content/12/en/part12b.md
```diff
@@ -89,7 +89,8 @@ Containerizing that should be relatively easy based on the previous example.
 - Use node as base
 - Set working directory so we don't interfere with the contents of the base image
 - Copy ALL of the files in this directory to the image
-- Start with DEBUG=playground:* npm start
+- Start with DEBUG=playground:* npm startaterial
```
mluukkai, line 92: this introduces a typo

src/content/12/en/part12b.md
```diff
@@ -190,7 +191,7 @@ RUN npm ci # highlight-line
 CMD DEBUG=playground:* npm start
 ```
 
-Even better, we can use _npm ci --only-production_ to not waste time installing development dependencies.
+Even better, we can use _npm ci --only=production_ to not waste time installing development dependencies.
```
mluukkai, line 194: hmm. both work. dunno which is the better command

mluukkai (changes requested):

PR #1523, opened 2021-10-08
Put request instead of get

tommybickerdike:
Underneath it says:
```
When we try to change the importance of the hardcoded note, we see the following error message in the console. The error says that the backend server responded to our HTTP PUT request with a status code 404 not found.
```
I believe this code sample should have been `.put`


mluukkai:
no, this is right

PR #1549, opened 2021-10-22
Update part8c.md

travisddraper:
The comparison for ObjectIds in the addAsFriend resolver won't function as expected without adjusting the used method (in my experience).


mluukkai:
fixed by another commit

PR #1569, opened 2021-11-03
Add hyperlink to combined reducer

akirataguchi115:


mluukkai:
in general I'm in favor of relative links

PR #1572, opened 2021-11-03
replace PUT with PATCH

keijop:
The assignment 2.18 is to update only phone number. Would it be more appropriate to use PATCH request to modify the resource instead of PUT to replace the whole resource?


mluukkai:
it would but this change would break too many things so will not do it now

keijop:
Yes, I saw chapter 3 continues developing backend for the same app so indeed it would cause problems down the line.

PR #1573, opened 2021-11-03
docker diff?

VincentLeV:
I ran container diff but it was not executed. I checked and I think the correct command would be "docker diff" instead of "container diff"


Jakousa:
Complete version should be "docker container diff". Same as the example above with `commit` the docker is omitted since we are talking about "Other docker **commands**".

```
$ docker --help

Usage:  docker COMMAND
```

If it's too confusing lets add docker at the start so it's the complete version with docker + command.

PR #1588, opened 2021-11-11
Update recommendation for mitigating SQL injection

PurkkaKoodari:
The current material says that you should escape user input before putting it in a SQL string.

## WRONG. You should basically never put user input in a SQL string.

- The list of people that tried to write SQL sanitation functions, and failed miserably, is long and contains many projects you've heard of. (The list of ways sanitation can go wrong is equally long and contains many things you can't come up with.)
- Parameterized queries were the suggested solution already [in 2008](https://stackoverflow.com/questions/60174/how-can-i-prevent-sql-injection-in-php). Basically every SQL library supports parameterized queries. If yours doesn't, it's far beyond time to upgrade.
- With parameterized queries you have to _try_ to get a SQL injection. And they might even be faster.

See e.g. [this Security.SE post](https://security.stackexchange.com/questions/230211/why-are-stored-procedures-and-prepared-statements-the-preferred-modern-methods-f) (which I'm suggesting for the new reference link), or the _many_ others available, for more explanation. 

This course is, admittedly, not about security, but at least it shouldn't teach insecure methods.

### DIFF
--- a/src/content/7/en/part7e.md
+++ b/src/content/7/en/part7e.md
@@ -313,7 +313,12 @@ SELECT * FROM Users WHERE name = 'Arto Hell-as'; DROP TABLE Users; --'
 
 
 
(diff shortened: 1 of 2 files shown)


mluukkai:
good point

PR #1592, opened 2021-11-13
Part 8 English proofreading

FlyingMandarine:
As with previous parts, a few grammatical mistakes and formatting errors have been fixed; overall, most of the fixes are minor and concern incorrect usage of commas and dashes.


lorenzojkrl (approved):

src/content/8/en/part8c.md
```diff
@@ -57,7 +57,7 @@ module.exports = mongoose.model('Person', schema)
 ```
 
 
-We also included a few validations. _required: true_, which ensures that value exists, which is actually redundant by using GraphQL we ensure that the fields exist. However it is good to also keep validation in the database. 
+We also included a few validations. _required: true_, which makes sures that a value exists, is actually redundant: we already ensure that the fields exist with GraphQL. However, it is good to also keep validation in the database.
```
lorenzojkrl, line 60: Which make sure

lorenzojkrl (changes requested):
Just a typo

lorenzojkrl (approved):

src/content/8/en/part8c.md
FlyingMandarine: Well-spotted! Fixed it; sorry about that. :)

PR #1602, opened 2021-11-16
Bug fix in use of useImperativeHandle

jeansibelius:
When `useImperativeHandle` is in the form as below, the ref will be callable in noteFormRef.current() (as stated [here](https://reactjs.org/docs/hooks-reference.html#useref))  

```
 useImperativeHandle(ref, () => {
  return {
    toggleVisibility
    }
  })
```

In order to make it callable with a custom name, the block should be defined as follows (as per [here](https://reactjs.org/docs/hooks-reference.html#useimperativehandle))

```
 useImperativeHandle(ref, () => ({
  toggle: () => {
    toggleVisibility()
    }
  }))
```

This can then be called with noteFormRef.current.toggle().

Proposing to simplify and stick with current definition and changing usage of `noteFormRef.current.toggleVisibility()` to `noteFormRef.current()`

### DIFF
--- a/src/content/5/en/part5b.md
+++ b/src/content/5/en/part5b.md
@@ -471,7 +471,7 @@ We can now hide the form by calling <i>noteFormRef.current.toggleVisibility()</i
 const App = () => {
   // ...
   const addNote = (noteObject) => {
-    noteFormRef.current.toggleVisibility() // highlight-line
+    noteFormRef.current() // highlight-line
     noteService
       .create(noteObject)
       .then(returnedNote => {     


mluukkai:
the proposed change does not work since noteFormRef.current() is an object, unless you define

```
 useImperativeHandle(ref, () => {
    return  toggleVisibility
  })
```

that would be a bit confusing

jeansibelius:
Thank you for reviewing this!

Agreed, looking at this now, what threw me off was assigning `toggleVisibility` function through the shorthand and adding the return explicitly to it compared to the [React documentation of useImperativeHandle](https://reactjs.org/docs/hooks-reference.html#useimperativehandle).

```
const toggleVisibility = () => {
    setVisible(!visible)
  }

useImperativeHandle(ref, () => {
  return {
    toggleVisibility
    }
  })
```

instead of this
```
const toggleVisibility = () => {
    setVisible(!visible)
  }

useImperativeHandle(ref, () => ({
    toggleVisibility
  })
```

If I now understand correctly, the same thing could also be written in the form of how it's written in the React documentation (doesn't make sense in this context, as `const toggleVisibility` is used by others as well, but just to clarify):
```
useImperativeHandle(ref, () => ({
    toggleVisibility: () => {
        setVisible(!visible)
    }
  })
```

PR #1616, opened 2021-11-22
Lock react-router-dom to 5.2.1 until part gets updated

otahontas:
Tested manually with newest cra, should work


mluukkai:
fixed in 2022 edition

PR #1636, opened 2021-11-26
Remove Requiring of the http module

andrew-yang:
Don't need it since we require express I think? http just causes some confusion especially since the recommendation is to use express.


mluukkai:
it is needed in this case

PR #1638, opened 2021-11-26
Fix missing 11.13

t-miller:
- Decremented all numbers starting at 11.14


mluukkai:
this change can no be made before the course break

PR #1647, opened 2021-11-30
Remove reference to master branch

parttis:
Remove reference to master branch because it does not exist.
Instead, the default branch in the repository is part3-1.


lorenzojkrl (commented):
I think the line that you are removing is adding some clarity. Any specific reason to remove it?

parttis:
> I think the line that you are removing is adding some clarity. Any specific reason to remove it?

There is no such branch as master in the repository in question. 

Instead of master, the name of the default branch is part3-1.
Here are all the branches for the repository listed if you want to check:
https://github.com/FullStack-HY/part3-notes-backend/branches/all

lorenzojkrl (approved):

PR #1656, opened 2021-12-05
More small updates to osa3a.md

parttis:
- Fix typos.
- Add dollar symbols to all shell command examples.


lorenzojkrl (commented):
Is the dollar sign an expected pattern throughout the whole site?

parttis:
> Is the dollar sign an expected pattern throughout the whole site?

In the beginning of the course, dollar signs were used consistently to indicate all the shell commands and make explicit distinction between command line use and code examples. But then at some part of the course, they were dropped out.

Either way is okay for me. It would be good to be consistent here.

lorenzojkrl (approved):

PR #1686, opened 2021-12-12
Import to run GraphQL playground

lorenzojkrl:
Without the following import
```
const {
  ApolloServerPluginLandingPageGraphQLPlayground
} = require("apollo-server-core");
```
users will get the following error: `ReferenceError: ApolloServerPluginLandingPageGraphQLPlayground is not defined.`
As reported in [apollographql.com](https://www.apollographql.com/docs/apollo-server/api/plugin/landing-pages/#graphql-playground-landing-page) "The GraphQL Playground plugin is not installed by default. To install it, import it from apollo-server-core and provide it to the ApolloServer constructor"

### DIFF
--- a/src/content/8/en/part8a.md
+++ b/src/content/8/en/part8a.md
@@ -275,7 +275,10 @@ The initial code is as follows:
 
 ```js
 const { ApolloServer, gql } = require('apollo-server')
-
+const {
+  ApolloServerPluginLandingPageGraphQLPlayground
+} = require("apollo-server-core");
+  
 let persons = [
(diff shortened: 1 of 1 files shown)


akirataguchi115 (changes requested):
Mine's looking fine without the plugin installed explicitly:
image

Do you have some sort of other error occurring without the plugin explicitly installed or is it just a warning?

mluukkai:
Switched from Playground to Apollo Studio Explorer in 2022 edition

PR #1690, opened 2021-12-16
Update providing app name to heroku as option bcoz its compusory!

sahilrajput03:
- Add showing the passing of _app-name_ as option for the heroku cli.
- Add the host as indicator because its differenly generated for users (NOT SAME FOR ALL AS INDICATED IN MATERIAL)>

### DIFF
--- a/src/content/13/en/part13a.md
+++ b/src/content/13/en/part13a.md
@@ -132,7 +132,7 @@ username=> \d
 (2 rows)
 ```
 
-In addition to the <i>notes</i> table, Postgres created a subtable called <i>not\_id\_seq</i>, which keeps track of what value is assigned to the <i>id</i> column when creating the next note.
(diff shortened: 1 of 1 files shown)


mluukkai:
hmm. worked for me without the app name 🤔

sahilrajput03:
@mluukkai https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/runs/4591623108?check_suite_focus=true

the pr didn't succeed for some reason... what could be the possible issue.., I tried to read the logs but it doens't make any sense to me. Do you understand that ?

mluukkai:
The content is merged and also live. The built fails if I merge two pull
requests within a too short time.

ke 22. jouluk. 2021 klo 19.21 Sahil Rajput ***@***.***)
kirjoitti:

PR #1695, opened 2021-12-21
Update part3d.md

GiorgiSheklashvili:


mluukkai:
this is correct already

PR #1699, opened 2021-12-24
Change the useEffect second parameter

matijanjic:
Changed the second parameter of useEffect from result to result.data.
Maybe I am in the wrong, but I couldn't get it to return to the all persons view without it.
It would, actually, but only for a fraction of the second and then it would switch back to the single person view.
Would love to hear the explanation for that!


mluukkai:
you are right. something has changed and it is now needed. changed in another commit

PR #1708, opened 2022-01-01
Update part4b.md - correction to promiseArray of noteObjects

josephalbaph:
const promiseArray = noteObjects.map(note => note.save()) can be wrong because it does not produces a promiseArray.
async and await need to be used by the map method to produce a promiseArray
suggested correction:
const promiseArray = noteObjects.map(async note => await note.save())


Kaltsoon:
`const promiseArray = noteObjects.map(note => note.save())` is correct, it returns an array of promises.

PR #1709, opened 2022-01-01
Fix typo line 418

ruile:
change returnedObject.id to returnedObject._id


ruile:
Apologies, the original is actually correct. While making this change will hide the **id** field when accessing /api/notes which is intended behaviour, it would cause the put and delete endpoints to fail with ```error message Cast to ObjectId failed for value "undefined" (type string) at path "_id" for model``` which is unintended behaviour.

I will close this pull request.

PR #1727, opened 2022-01-10
Update part3c.md

ihamuma:
Add information on fixing Heroku-MongoDB integration while using dotenv.


mluukkai:
one should not set the PORT it is provided by heroku

mluukkai:
fixed in another commit

PR #1731, opened 2022-01-13
Add note about WSL2 + Cypress

### LINKED ISSUE #1729: Part 5d - Using Cypress with WSL2
Users that are using WSL2 for completing exercises must perform some additional configuration to be able to use Cypress and finish exercises from 5.17. to 5.22.

Cypress docs on how to set up WSL2 to be able to use Cypress - [https://docs.cypress.io/guides/getting-started/installing-cypress#Windows-Subsystem-for-Linux](https://docs.cypress.io/guides/getting-started/installing-cypress#Windows-Subsystem-for-Linux)
A blog post on how to do the set up - [https://nickymeuleman.netlify.app/blog/gui-on-wsl2-cypress](https://nickymeuleman.netlify.app/blog/gui-on-wsl2-cypress)

The second link (blog post) is more in depth, because for Cypress to work with WSL2 you need to install some additional dependencies to Linux. I set up Cypress on my WSL2 by using both guide (obviously they overlap with the same instructions for the most part).

I think that it would be useful to add a simple heads up message to the Cypress part to just notify people that if they are using WSL2, they will need to do some extra steps before they can use it and provide either one or both of this two links, which explain how to perform the set up. Like there is a message about potential issues with the updated mongoose-unique-validator in part 4c, there could be a similar message about potential issues with Cypress, if you are developing with WSL2.

akirataguchi115:
Closes #1729

### DIFF
--- a/src/content/5/en/part5d.md
+++ b/src/content/5/en/part5d.md
@@ -91,6 +91,7 @@ Let's add an npm-script to <i>the backend</i> which starts it in test mode, or s
 ```
 
 <!-- Kun backend ja frontend ovat käynnissä, voidaan käynnistää Cypress komennolla -->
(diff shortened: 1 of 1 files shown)


mluukkai:
fixed in another commit

PR #1732, opened 2022-01-13
part13c: Fix grammar, typos

protofarer:
1. logging *in* instead of just logging (easily confused with logging to stdout)
2. for section *admin user and user disabling*, code block for file 'util/db.js': 
a. fix inconsistent reference to function "rollbackMigrations" to "rollbackMigration"
b. line429: fix import statement from 'move' to 'Umzug' or 'umzug'
c. line468: replace 'migration' with 'Umzug'
d. line477: replace 'move' with 'Umzug'


Kaltsoon:
Thanks, these should be fixed now!

PR #1744, opened 2022-01-16
Update part2d.md

ricardo-afonso:
Remove using the index as key - Described in an earlier part as a practice to avoid. Using the `note.id` property in place.


mluukkai:
changed by another commit

PR #1747, opened 2022-01-19
Fix typos, improved clarity, removed unnecessary repetition part12c

jamessl154:
Set up is the verb form
https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/s/set-up-setup

Heavyweight "a person or thing that is important or serious and that other people notice" https://dictionary.cambridge.org/dictionary/english/heavyweight

Most of the changes I understand what you're trying to say but I think it always helps to be more clear

e.g. todo-frontend and todo-backend are more clear than todo-front and todo-back

line 588 is a repetition of line 575 which I think is not needed

Also I'm unsure if line 531 is redundant


akirataguchi115 (approved):
LGTM!

PR #1751, opened 2022-01-22
Update part1b.md

flaviaouyang:
Fixed two sentences that are unnecessarily grammatically complex.
Add JavaScript to the first sentence in the Class section so that it's clear which programming language it's referring to.


mluukkai:
fixed in another commit

PR #1756, opened 2022-01-26
Fix typo on Finnish translation

FinThunderstorm:
yliopist => yliopisto


mluukkai:
fixed in another commit

PR #1758, opened 2022-01-26
Update part0a.md

aalisina:
I think the submission dates for the year 2022 could be updated.


aalisina:
I think the submission dates for the year 2022 could be updated.

mluukkai:
will be updated in another commit

PR #1763, opened 2022-01-30
Part 9c updated /ping to /api/ping

coolbigbear:
Pationator app pings at /api/ping not /ping


mluukkai:
fixed in another commit

PR #1764, opened 2022-01-30
Replace mongo mock with mongodb-memory-server

kazumaneet:
The documentation for mongo mock isn't very descriptive so setting it up is harder. However, I've personally used mongodb-memory-server which is very well documented so is much easier to set up.


mluukkai:
Thanks for suggestion! Included now in material by another commit.

PR #1767, opened 2022-02-03
Changed "markup" to "stylesheet" under chapter CSS

sheikhameen:
Changed "CSS, is a markup language"  to "CSS, is a style sheet language". CSS is **NOT** a markup language.
https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/CSS_basics#what_is_css


mluukkai:
changed in 2022 version by another commit

PR #1784, opened 2022-02-15
Update part8a.md to include GraphQL Playground

CaprioZero:
As GraphQL Playground is deprecate in latest Apollo 3, so we need to include import plugin to be able to use Playground in localhost and not need to use Apollo online services. Also restructure object within object part to be more linear


mluukkai:
we have switched from playground to apollo explorer

PR #1785, opened 2022-02-15
Remove deprecation warnings

akirataguchi115:
Changing the code to match the [actual code in GitHub](https://github.com/FullStack-HY/graphql-phonebook-backend/blob/part8-4/index.js). For more information check https://mongoosejs.com/docs/5.x/docs/deprecations.html :monkey:


mluukkai:
don't know why that was missing

PR #1788, opened 2022-02-16
Fix: failing cypress tests

magnusws:
I found that several of the test examples in the course material fails. When running these tests cypress
actually required the show all button to be clicked for the '.contains' function to be able to access the notes that are defined as not important.


mluukkai:
fixed in another commit

PR #1806, opened 2022-02-22
Update part0b.md

Syhner:
Minor grammar correction (by -> when)


Syhner:
> Create a similar diagram depicting the situation where the user creates a new note on page https://studies.cs.helsinki.fi/exampleapp/notes by writing something into the text field and clicking the submit button. 

"by" can imply that creating the diagram is done through "writing something into the text field and clicking the submit button".

"when" would be more a more succinct choice since it removes this vagueness.

PR #1817, opened 2022-02-26
Update part4: regarding mongoose-validator

mariano-farace:
For what it's worth, I have "mongoose": "^6.0.6", "mongoose-unique-validator": "^2.0.1" and it seems to be working properly, instead of using "mongoose-unique-validator": "3.x".  PD: First ever pull request on a real project.


mluukkai:
fixed in another commit

PR #1825, opened 2022-03-03
Update part3b.md

yaakovab:
I deleted the lines in the head of this article that was about making a change also in the Effect in the app.js file because at the end of part 2 we transferred all activity of "server" to the services/notes/js file, and from that file imported getAll function that gets all the notes for us, and in file app.js there was no more any URL. so, there is nothing to change.


mluukkai:
fixed by another commit

PR #1839, opened 2022-03-09
Part2d: Fix eslint warning

tonilaukka:
Current version of create-react-app and eslint config will give a warning:
`Assign object to a variable before exporting as module default eslint[import/no-anonymous-default-export](https://github.com/import-js/eslint-plugin-import/blob/v2.25.4/docs/rules/no-anonymous-default-export.md)`


mluukkai:
will not fix this. it is in many places of the material and also in code, so fixing just one place is not enough, we will start using eslint in part 5 and the rule is not enforced then

PR #1847, opened 2022-03-11
Update part5a.md

yashmeen:
fixed (I think?) to reflect what the user is seeing at this point in their code editor


mluukkai:
fixed by another commit

PR #1893, opened 2022-03-25
options usefindandmodify, usecreateindex are not supported

enkeipy:
1) The config for connection is misplaced - we just log it instead of using it.
2) But even if use it like mongoose.connect(MONGODB_URI, {
    useNewUrlParser: true,
    useUnifiedTopology: true,
    useFindAndModify: false,
    useCreateIndex: true
  })
it gives following error message ("mongoose": "^6.2.8"): options usefindandmodify, usecreateindex are not supported.
So maybe these are deprecated. Project works with just plain mongoose.connect(MONGODB_URI), but it also doesn't give any warning for using useNewUrlParser and useUnifiedTopology. Thanks!

### DIFF
--- a/src/content/8/en/part8c.md
+++ b/src/content/8/en/part8c.md
@@ -61,9 +61,9 @@ const Person = require('./models/person')
 
 const MONGODB_URI = 'mongodb+srv://databaseurlhere'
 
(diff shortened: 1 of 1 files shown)


mluukkai:
shit, no idea why it was only in log

PR #1918, opened 2022-04-06
Update part1a.md

ThePiyushAggarwal:
updated index.js to react 18


mluukkai:
changed in another commit

PR #1942, opened 2022-04-14
Update part1a.md

sailingnn:
Added index.js for React version 18 to prevent a warning on the browser console


mluukkai:
this is already in react 18 format

PR #1955, opened 2022-04-20
Update part2c.md

sailingnn:
updates the address of the endpoint "all" of restcountries.com


sailingnn:
remove the single quotation marks when set REACT_APP_API_KEY so that users do not have to deal with them in App.js

PR #1961, opened 2022-04-21
Updated material and added fixes for user-event 14

ishaanb5:
## Premise
The current tests in the course material for part-5c are failing due to the [breaking changes](https://github.com/testing-library/user-event/releases/tag/v14.0.0) in user-event@14. The change that impacts the test cases is:  APIs always return promise. 

## Changes
- Updated links referring to previous version of user-event.
- Added description for starting a user session
- Added call for starting a user session in tests and made calls to user-event APIs asynchronous.


Kaltsoon:
Awesome stuff, thanks!

PR #1970, opened 2022-04-23
Update part5a.md

kkiilas:
The links in the subsection "Exercises 5.1-5.4" and in exercise 5.1 point to different repositories with the same contents. I would suggest using https://github.com/FullStack-HY/bloglist-frontend in both places because it has been forked four more times than the other repository.


mluukkai:
https://github.com/FullStack-HY2020/bloglist-frontend is for 2022 and https://github.com/FullStack-HY/bloglist-frontend was last years version

kkiilas:
Hyvä tietää, kiitos.

PR #1995, opened 2022-05-01
It may be more appropriate to use modal.

ZZZCNY:


Kaltsoon:
Why?

ZZZCNY:
> Why?

The Menu will pop up around the parent element instead of being centered on the screen.

Kaltsoon:
Is that a problem?

> Menus display a list of choices on temporary elevated surfaces. Their placement varies based on the element that opens them.

Sounds suitable for the use case.

ZZZCNY:
At least in the screenshot from Exercise 10.23, it's centered.

PR #2001, opened 2022-05-07
Fixing imported redux action

edumt:
"initializeNotes" is introduced only in "Asynchronous actions and redux thunk" section, the imported action should be "setNotes" as used inside the useEffect hook or in the index.js implementation example (lines 195 to 212 in the source code).


Kaltsoon:
This seems to be fixed already. Thanks!

PR #2016, opened 2022-05-16
Update part3c.md

believerd:
typo fix
URI -> URL


mluukkai:
it is written URI in many places so can not just change from one. and URI is not a typo, it is intentional https://webmasters.stackexchange.com/questions/19101/what-is-the-difference-between-a-uri-and-a-url

but URL would also do fine.

PR #2027, opened 2022-05-25
Fix typo in credit requirements

PremSagarS:
The requirements for getting credits in Part 6 should be completion of exercises in parts 0-6 and not parts 0-7


mluukkai:
this is not a typo

PR #2031, opened 2022-05-26
Change structure of mongo.js

gkampiotis:
The original code throws connection error. Most probably it is because it tries to save the note before the connection is established.

By placing the Schema and note/note.save() inside mongoose.connect(url) it is now possible to post the note when the connection is for sure established.


Kaltsoon:
Thanks!

PR #2035, opened 2022-05-29
Update part4a.md

edgarfb:
There is not need to create a server because app itself return a server


mluukkai:
clarified by another commit

PR #2045, opened 2022-06-02
part13c | change snake case in Membership Model to camel case

prb01:
Model example Membership for many-many relationships shows user_id & team_id instead of userId & teamId.


mluukkai:
fixed by another commit

PR #2049, opened 2022-06-05
Update part12b.md

Grazvis1988:
Found a typo(line 559).And updated to new version of docker compose. Now instead of "docker-compose" docker uses "docker compose"(v2.5.0)...


mluukkai:
the use of docker-compose is intentonal, so will not merge this yet @Jakousa pls what is your opinion

mluukkai:
explanation about docker compose vs docker-compose added

PR #2057, opened 2022-06-14
add missing ApolloClient

buiquanghuy23103:
The code that updates cache inside useSubscription needs ApolloClient, which is missing.


mluukkai:
client is there, see https://github.com/fullstack-hy2020/graphql-phonebook-frontend/blob/part8-9/src/App.js#L38

PR #2072, opened 2022-06-21
Fix missing default case i.e, when important query parameter is not supplied.

sahilrajput03:


mluukkai:
do not believe this is needed

PR #2079, opened 2022-06-23
Replace with much better stackoverflow answer for part 1d

RohanSreelesh:
Much better stackoverflow answer for why state shouldn't be mutated directly. This helped me since the original answer linked in the content didn't clarify it well in my opinion.

here is the link: https://stackoverflow.com/a/40309023

### DIFF
--- a/src/content/1/en/part1d.md
+++ b/src/content/1/en/part1d.md
@@ -197,7 +197,7 @@ const handleLeftClick = () => {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sahilrajput03:
I feel same, your stackoverflow answer is much more relevant to the basic simple question "why can't we mutate state". The older stackoverflow answers a different question i.e, why not just shallow clone the state to trigger a state change.

I support your this PR. Btw I added the current stackoverflow answer earlier via PR https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/pull/1991.

PR #2083, opened 2022-06-26
Requires camelCase instead of snake_case.

sahilrajput03:
In model definition of membership model it requires camelCase instead of snake_case.


sahilrajput03:
Reference:

image

PR #2095, opened 2022-07-09
Update part0b.md

CarlosFSousa:
Changing from getElementById to getElementsByClassName.
The attribute that was given was class 'notes' and not id.


CarlosFSousa:
I see that it repeats two more times. I can change those as well.

PR #2109, opened 2022-07-13
Added some punctuation to the create git repo part

WillTrem:
I felt some punctation was missing from that part, which made it harder to read!


gjohnsx:
I feel that adding the punctuation is not necessary in an unordered list. In addition, it might confuse people who try to add the period to the command line.

PR #2110, opened 2022-07-13
part6c - replaced 'such that' with 'so that'

faisne:
'so that' is more appropriate in given context

### DIFF
--- a/src/content/6/en/part6c.md
+++ b/src/content/6/en/part6c.md
@@ -7,7 +7,7 @@ lang: en
 
 <div class="content">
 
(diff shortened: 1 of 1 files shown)


gjohnsx:
I think 'such that' is fine here, but the comma should be removed.

"Let's expand the application such that the notes are stored to the backend" reads more naturally.

faisne:
I'm not a native English speaker, so I'm not 100% sure, but I think 'such' answers 'which' or 'what' and should correspond to a noun (e.g. 'such thing', 'such a way', and the latter would be ok here instead of just 'such'), and 'so' answers 'how', 'in what way' and can correspond to a whole clause, which is the case here ('the notes are...'). 

I found this link that supports this point of view https://dictionary.cambridge.org/grammar/british-grammar/such-or-so

PR #2176, opened 2022-08-04
Would advise updating pictures

IanHefflefinger:
Some of the pictures used on. this page are from when this project was on Heroku. The project now exists on the helsinki.fi website and the rest of the content reflects that. It is a minor change, but could assist those who could be confused about the content vs the screenshots.

### DIFF
--- a/src/content/0/en/part0b.md
+++ b/src/content/0/en/part0b.md
@@ -5,6 +5,8 @@ letter: b
 lang: en
 ---
 
+<!-- update photos -->
+
 <div class="content">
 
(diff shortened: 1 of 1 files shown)


rikurauhala (approved):

rikurauhala:
I've noticed similar issues in some parts of the course. For example, in part9 the [screenshots](https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/blob/source/src/content/images/9/35c.png) may contain methods or syntax (such as `React.FC`) that are no longer discussed in the course material.

Some pictures should definitely be updated. Screenshots should match what is actually written in the course material.

Perhaps consider opening a new issue related to this?

PR #2192, opened 2022-08-16
Update part3a.md

Medmly20208:


Medmly20208:
adding notes type in DELETE method part in this case 'const'

PR #2204, opened 2022-08-23
Update part11b.md

edward-designer:
I am using Cypress 10.6, the originally suggested steps for testing just got stuck. After changing to cypress-io/github-action@v4 and taking out command: npm run test:e2e, the tests are performed as expected.


mtuomiko:
Seems peculiar, no such issues here. Are you sure have/had a suitable npm script defined, that is, it's named `test:e2e` (or whatever you refer to in the pipeline config) with command `cypress run` (`--headless` argument seems to be default in Cypress v8 and higher)?

I mean `cypress-io/github-action` (at least v4?) seems to default to the same behavior when you don't override the command but it's not like the material is wrong here.

PR #2243, opened 2022-09-15
Update part4b.md -- Simplify Note.save() batch operation

steffennilsen:
Part 4B - Simplify

```js
let noteObject = new Note(initialNotes[0])
await noteObject.save()
noteObject = new Note(initialNotes[1])
await noteObject.save()
```

to 
```js
await Note.collection.insertMany(initialNotes)
```

### DIFF
--- a/src/content/4/en/part4b.md
+++ b/src/content/4/en/part4b.md
@@ -321,18 +321,13 @@ const initialNotes = [
 // highlight-start
 beforeEach(async () => {
   await Note.deleteMany({})
-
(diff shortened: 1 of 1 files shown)


steffennilsen:
There's a whole paragraph about `Optimizing the beforeEach function` further down the file advocating to use `Promise.all()` instead. I doubt removing the earlier unoptimized code is suitable then. Maybe introduce as a line after going over `.all()` or just not worth mentioning `Model.collection.insertMany`?

PR #2248, opened 2022-09-16
Update the links to TS utility types

erictelkkala:
The format of section has changed, so the links didn't scroll to the correct section when clicked.


erictelkkala:
The links were previously plain HTTP links, but the new ones are HTTPS

PR #2251, opened 2022-09-18
Update part12b.md

Uimalakki:
mongo shell is removed from MongoDB 6.0. The replacement is mongosh


Uimalakki:
Already pull request with the same issue.

PR #2262, opened 2022-09-26
Grammatical Corrections part3b.md

AKMalice:
Grammatical Corrections in Part 3b

### DIFF
--- a/src/content/3/en/part3b.md
+++ b/src/content/3/en/part3b.md
@@ -76,13 +76,13 @@ The react app running in the browser now fetches the data from node/express-serv
 
 Now that the whole stack is ready, let's move our application to the internet.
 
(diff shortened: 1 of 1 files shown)


src/content/3/en/part3b.md
```diff
@@ -101,23 +101,23 @@ If you decide to use [Fly.io](https://fly.io/) begin by installing their flyctl
...
+*Note* if the command _fly_ does not work on your machine, you can try the longer version _flyctl_. Eg. on MacOS, both forms of the command work.
 
 Initializing an app happens by running the following command in the root directory of the app
 
 ```bash
 fly launch
 ```
 
-Give the app a name or let Fly.io autogenerate one. Pick a region where the app will be run. Do not create a postgres database for the app since it is not needed.
+Give the app a name or let Fly.io auto generate one. Pick a region where the app will be run. Do not create a postgress database for the app since it is not needed.
```
rikurauhala, line 120: I believe "postgres" refers to [PostgreSQL](https://www.postgresql.org/) so no error here?

PR #2275, opened 2022-10-01
Added German Translation

NicolasWebDeveloper:


laurin-no:
I took a quick look and I have to say that the PR definitely needs improvement. The first major issue I see is the mixed use of "du" and "Sie" (informal and formal you) in the challenge description.

Furthermore some of the translation strings sound (or better read) like they have been translated word by word, which sound quite awkward to me as a German native speaker. One example is the translation string "Das ist die Full Stack Herausforderung", which I admit is hard to translate but this translation feels very clumsy and I would not translate "challenge" to "Herausforderung", since "challenge" is widely used in German.

Another example is the translation "nextPart" to "Nächster Part", which is the other extreme in comparison to the previous example because "part" should be translated to something like "Kapitel" and not "Part", which is something different according to the Duden (one of the main German dictionaries): [https://www.duden.de/rechtschreibung/Part](https://www.duden.de/rechtschreibung/Part).

To summarize my comment, this PR needs to be reworked from the ground up because of the poor translation quality.

PR #2283, opened 2022-10-04
Corrected Text element

erictelkkala:
Seems to be not relevant, or just incorrect element in the code sample, since the project doesn't require the students to create a custom Text element.

Should be imported from 'react-native'.


erictelkkala:
Ah never mind, this was actually talked about in [https://fullstackopen.com/en/part10/react_native_basics#consistent-user-interface-with-theming](https://fullstackopen.com/en/part10/react_native_basics#consistent-user-interface-with-theming)

PR #2288, opened 2022-10-08
Specify which application

IanHefflefinger:
It is assumed this means the phonebook application, but it would be best to specify here as the tutorial takes the user through the notes application and could lead to confusion over which one.


mluukkai (changes requested):
It is the phonebook, same that you have been working with in all the assignments, can not obviously merge this as it is

PR #2289, opened 2022-10-08
Improve accessibility

peippo:
Add various accessibility improvements

- Improve semantic structure by using `header`, `main` and `footer`
- Add "skip to content" link for keyboard / screen reader users
- Add visual indicator for active main navigation link [1]
- Add labels for search page input & language picker select
- Add Accordion component ARIA expanded & controls attributes, hide closed panel from DOM and add keyboard focus styling
- Add Content liftup component focus styling & heading element

[1] Course contents page does not get set as active due to it being just an anchor


Kaltsoon:
Good stuff, thanks!

PR #2304, opened 2022-10-18
Clarify the setup of thunk

laurin-no:
According to the [installation guide](https://github.com/reduxjs/redux-thunk#installation-and-setup) we do not need to run `npm install redux-thunk`, so I suggest removing it and mentioning the possibility of setting up the library manually and an example case, when this process can make sense.


mluukkai:
done in another commit

PR #2324, opened 2022-10-27
change ClassName to Id

adeleke5140:
the code section depicted used the `document.getElementsByClassName` DOM API to get the `element` with the class of `notes` but no such elements exists. It should rather be the id of `notes`


mluukkai:
that is not true

adeleke5140:
> that is not true

What do you mean?
Is it that the error I pointed out wasn't present?

I see it has been updated.

PR #2340, opened 2022-11-06
Update part2a.md - display correct object error

charliedodds:
Update concatenating a string with an object to reflect correct output


charliedodds:
didn't see this one before or would have proposed them together

PR #2343, opened 2022-11-10
Update part3b.md

mishrasur7:
In line 98, about the configuration of port for Fly.io or Heroku, I tried the port number without double quotes but it did not work for me, as I tried couple of times. But when I placed the port number inside double quotes then it started working. Is there anything that I have misunderstood or does the port number should be inside double quotes?


mluukkai:
when deploying to fly, the code uses process.env.PORT so the problem must be elsewhere 🤔 

I could look for the issue if you could eg email me your repository

mishrasur7:
Thank you for your reply. 

I have send you an email mentioning by repos in the address mluukkai@iki.fi

PR #2355, opened 2022-11-17
Update part3d.md

mishrasur7:
When I installed eslint and configured, the configuration is saved in a file named .eslintrc.cjs but according to your instruction, configurations are saved in a file named .eslintrc.js What is the difference between these two extensions of .cjs and .js or does the instruction on material needs to be modified?

### DIFF
--- a/src/content/3/en/part3d.md
+++ b/src/content/3/en/part3d.md
@@ -231,7 +231,7 @@ We will answer all of the questions:
 
 ![](../../images/3/52be.png)
 
-The configuration will be saved in the _.eslintrc.js_ file:
+The configuration will be saved in the _.eslintrc.cjs_ file:
 
 ```js
 module.exports = {


snowbytes:
It has to do with how Node determines which module system it will use. `.cjs` is the extension to explicitly use commonJS modules (that use `require`) whereas `.mjs` is the one for ESM modules (that use `import`).

Node projects use commonJS by default for backwards compatibility, but can and should specify which system they use globally by adding a `"type": "module"` or `"type": "commonJS"` in their `package.json`.

Since external tools such as project or config generators don't know which system your project uses globally, they often opt to use `.cjs` or `.mjs` instead. For mor in-depth info, check [the official docs](https://nodejs.org/api/packages.html#determining-module-system)

mishrasur7:
@snowbytes Thank you for your explanations

PR #2359, opened 2022-11-18
Update part4b.md

mishrasur7:
I have some concern about the line 110 
When we define an url for TEST_MONGODB_URI with testNoteApp database name then initially it does not have any data to test and hence our test cases will fail. Is it possible to have some explanations on this for me?


mishrasur7:
My apologies for making changes to this and pull request. Initially I thought that the new database does not have any data to test that is why I made the change but now I understood that we have the beforeEach function to clear the new database and insert data into database before running each test. This also makes sure that we are not making any harm to our actual database that is in production already.

PR #2373, opened 2022-12-01
Update part1d.md

jobjer:
Should specify year to be more precise.


mluukkai:
this note was not anymore valid so it is now removed

PR #2397, opened 2022-12-22
Update part8a.md

mishrasur7:
Updating the code to use the latest version of Apollo server. 

I got some issues while using the version 3.10.1 of Apollo server and I followed the official documentation guidelines of Apollo server. If some body else is also facing same issue then, this update might help.


mluukkai:
course is still recommending the older apollo, will fix this later

PR #2398, opened 2022-12-22
Update part8a.md

mishrasur7:
Updating error handling with the latest version of Apollo server


mluukkai:
course is still recommending the older apollo, will fix this later

PR #2411, opened 2022-12-31
Autofix markdown lint parts 0-7

onzfonz:
This PR is to address some of the rules that are covered in [markdown lint](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md) that I think would be easy fixes though they require larger code changes.  For these changes, I used a [vscode Markdown Lint Plugin](https://github.com/DavidAnson/markdownlint) 

autofix (the changes related to this PR) - are mostly to fix rules around trailing spaces (having spaces at the end of lines that isn't a 0 or 2), putting blank lines around headers and code blocks, and providing syntax languages for code blocks.  I think a couple that I used (like plantuml in 0b) are not supported by gatsby, but wanted to use it anyways as I thought it would be nice to describe (it does throw a warning during gatsby develop, so happy to move out as well)

I also created another pull request #2412 that builds off of this one but makes more opinionated changes

and finally a [controversial branch](https://github.com/onzfonz/comp227/tree/controversial_markdown_lint) where I changed all underscores to asterisks for example and so I thought that is much more subjective.

The markdown lint config file that I used for all three is below:

```json
"markdownlint.config": {
            "default": true,
            "MD003": {
              "style": "atx" //atx is the current style used in the document
            },
            "MD004": {
              "style": "dash" //I use dashes as that is what github defaults to with unordered lists
            },
            "MD007": {
              "indent": 4 //again, completely subjective to use 4 spaces as indentation for unordered sublists
            },
            "MD013": false, //turns off max number of characters on a line and can be configured more
            "MD033": false, //turns off the use of inline HTML
            "MD035": {
                "style": "---" //style to use for horizontal rules
            },
            "MD041": false, //turns off rule about first line in file needing to be a top-level header
            "MD049": { //only followed for controversial branch
                "style":"asterisk"
            },
            "MD050": { //only followed for controversial branch
                "style":"asterisk"
            }
    },
```
The goal would be that if this ends up being followed for the entire project, perhaps a check could be instituted for future markdown changes

PR #2412, opened 2022-12-31
Changed headers and made more opinionated markdown lint choices

onzfonz:
This PR is based off of #2411 and includes commits that deal with what I describe as more opinionated choices.  Those commits have the commits prefix while the ones from #2411 have the autofix prefix

The reason for these fixes in this one is that there is a [rule MD024](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md024---multiple-headers-with-the-same-content) - multiple headers with the same content that I think should be followed in case one ever wants to link to that header in the future which happens in 7c  and I also followed
[MD026](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md026---trailing-punctuation-in-header)

As I noticed this one it largely changes all of the exercises to not have a dot at the end of them.  I think this rule could be changed based on the configuration to be false, but in this PR, it has the dots removed, which I think is fine to ignore or not change.  (If you'd like it to be reverted back, that is totally fine)


onzfonz:
Sorry, I think in this one I ended up doing some of the changes that I had in #2416 where I changed some of the headings from underscores to asterisks to enforce consistency.

PR #2419, opened 2023-01-04
updated to apollo 4

pratikdevdas:
The app was breaking and now it works it has been tested thorughly and carefully updated with help of apollo server docs.


mluukkai:

PR #2426, opened 2023-01-11
Part8a, updated import for Apollo Server v3

sheikhameen:
The `apollo-server` package is part of Apollo Server v2 and v3, and has to be imported/required from 'apollo-server', not '@apollo/server'. The latter is for version 4.


mluukkai:

PR #2428, opened 2023-01-12
Clarification

eemanioui:
The issue described in the lesson is not related to CORS, rather it has to do with the limitations imposed by the `same origin policy` implemented by browsers to prevent security vulnerabilities when it comes to issuing dynamic AJAX requests to URLs that don't share the same origin as the source URL.

CORS came as a solution to solve the limitation imposed by browsers when it comes to same origin policy. Therefore, it is a solution and not the core issue as alluded to by the section in this lesson.


mluukkai:
Thanks, this is a great addition!

PR #2430, opened 2023-01-13
Update part3b.md

minhle27:
For fly.io, the internal port should be 8080. The line of code: "const PORT = process.env.PORT || 3001" won't work for fly.io


mluukkai:
it works, the problem is elsewhere. the material has is showing the right trick

PR #2440, opened 2023-01-18
fix typo

kevin-weitgenant:


kevin-weitgenant (commented):
There are more occurrences I only noticed now. It's better to Ctrl H all of them

PR #2457, opened 2023-01-22
some minor fixes

leomfn:
* fix mongodb URI/template literal (maybe also add instructions on how to start the app without the password included in the file, e. g. calling `DB_PASSWORD=XXYYYZZZ npm start` and using `process.env.DB_PASSWORD` within URI template literal)
* add comma


leomfn:
Didn't see that dotenv explanation is included later in the chapter, so maybe just dismiss this hint

> maybe also add instructions on how to start the app without the password included in the file, e. g. calling `DB_PASSWORD=XXYYYZZZ npm start` and using `process.env.DB_PASSWORD` within URI template literal

PR #2466, opened 2023-01-25
Minor fixes in the attribute names

antoniolopez7217:
The attribute was called "onClick" in the previous definitions of the function "Button" (row 568, 571 and 586 for example), so it should keep the same name and not change it to "handleClick". Otherwise, the program will not work and clicking the buttons will have no effect on "counter"


mluukkai:
this is not quite the fix that is needed, there is now a fix in another commit

PR #2470, opened 2023-01-28
Fix broken links

kiliczsh:
@mluukkai for visibility


kiliczsh:
For part2a is previously created PR: #2469

PR #2476, opened 2023-01-30
updating apollo server docs

Naf256:
some of the code for graphql in part-8 was broken because of some change in version. so, they were updated accoding to latest apollo server documentation


mluukkai:

PR #2482, opened 2023-01-31
Update part3c.md

Jimreed91:
Perhaps I'm misinterpreting but if this sentence is referring to just allowing any user to make changes I think it should be structured like this


mluukkai:
there is actually a slight problem with translation. will fix this with another commit

PR #2491, opened 2023-02-01
fix: link on markdown

Dezzep:
When clicking the link from the full stack open website, (section 6.19. -6.21) the link directs me to https://fullstackopen.com/en/part6/(https://github.com/fullstack-hy2020/query-anecdotes). I believe this is happening because in the markdown, the link is wrapped in double parentheses [this project]((linkhere)).

I have removed the double parentheses to resolve this issue.


Dezzep:
Clicking on the link above (this project), replicates the issue. Because it is a markdown link wrapped in double parentheses `[thisproject]((linkhere))`

PR #2526, opened 2023-02-08
Update part5d.md

hanasiu:


hanasiu:
My native language is not English. That's why I wrote the description too sparsely. I don't know what it is on other OS. It would be nice if it is written in a decent form. etc) NB ... reference site.. 
[jmargolisvt](https://github.com/jmargolisvt) refers to mac os. Looks like some more research is needed

hanasiu:
It works. thank you very much.

PR #2533, opened 2023-02-10
Update part0b.md

oarchner:
Get element by id not by class


mluukkai:
it is class, not id:

image

PR #2563, opened 2023-02-13
Fix to tests failing in Part 4 - Error handling and async/await

P4ps:
Currently, line 755 fails the tests. Comparing resultNote.body to noteToView fails since resultNote.body.date is a date formatted as a String whereas noteToView.date is a Date object.

The solution is to simply use the date String in resultNote.body.date to create a new Date object and reassign to itself.

Part4_ErrorHandlingAndAsyncAwait


mluukkai:
hmm. the date field should have been removed, will check it tomorrow

mluukkai:
indeed, the date was still left in quite a many place in parts 3&4. now it is gone

PR #2597, opened 2023-02-19
better setVisible logic

sirbh:


vejol:
Thank you for the pull request. We believe the original code is clearer, so I will not accept this change at this time.

PR #2636, opened 2023-02-26
Update part1b.md

eMUQI:
add tranlation


eMUQI:
https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/blob/source/src/content/1/zh/part1b.md

When an <i></i> tag is included within a comment, the comment becomes invalid and the content within the comment may appear on the webpage. How can this be resolved? Please refer to line 307 337.

PR #2637, opened 2023-02-26
Fix typo change handleClick to onClick to match the component prop

cmd0x5f:
Current code with handleClick as prop doesn't work.


mluukkai:
fixed by another commit

PR #2662, opened 2023-03-02
Why BrowserRouter doesn't work when you refresh

keshabmanni:
Added info about BrowserRouter and why it doesn't work when you refresh and about HashRouter


mluukkai:
a new section in part 7 has the right code now

Issue #2661, opened 2023-03-02
Dark Mode: Bad Text Color


Juanescacha:
In dark mode, the text for the elements of the list of the info update of the parts intro is color white, should be black like in light mode, for good visibility.

Dark Mode
![image](https://user-images.githubusercontent.com/6394715/222353999-08c4d21c-ba02-468f-aaf1-e3c6a13da416.png)

Light Mode
![image](https://user-images.githubusercontent.com/6394715/222354040-260775aa-e090-43f5-aa60-e77aa0435bbe.png)

Juanescacha:
The suggested solution for this white text is in this PR #2663

Juanescacha:
Fixed

PR #2663, opened 2023-03-03
color fix of part updates on dark mode

Juanescacha:


Juanescacha:
Change for fixing this issue on dark mode #2661

PR #2703, opened 2023-03-08
Add React Hook Form section

matnreyes:
I think more and more people are moving onto React Hook Form. I wrote a very quick explanation as to how it works but I trust the FSO team are much better equipped at explaining this concept. From  what I've seen online, Formik has started to become abandoned and it's slowed and much heavier than RHF. Also, validation can be done straight from this single library using Regex. 

This video does an amazing explanation for the concept. I created a login form using RHF instead of Formik.

https://www.youtube.com/watch?v=G4jD_u7isXk


mluukkai:
Indeed Formik should be replaced with the React Hook Form. It shall happen sometime later, so I will not yet merge this PR

PR #2714, opened 2023-03-11
Name query "changeNumber" so it matches other code

frozenzia:
Also CHANGE_NUMBER instead of EDIT_NUMBER -- this is more in line with the previous example of CREATE_PERSON.


mluukkai:
well, merging this would put the example code in github to out of synch, and the same change should be done to all language versions so will do this later

PR #2715, opened 2023-03-11
Add explanation why event handler can be simplified

noelsantotomas:
Add explanation as to why "Assigning the object to a variable in the event handlers is not necessary"


mluukkai:
no, you have a misunderstanding here, this has nothing to do react merging the state. and with state hooks it is not actually happening anymore

PR #2725, opened 2023-03-13
Correcting small error in part1c.md

gurvesh:
Correct errata - handleClick prop, instead of onClick prop

### DIFF
--- a/src/content/1/en/part1c.md
+++ b/src/content/1/en/part1c.md
@@ -629,7 +629,7 @@ const App = () => {
 
 Since we now have an easily reusable <i>Button</i> component, we've also implemented new functionality into our application by adding a button that can be used to decrement the counter.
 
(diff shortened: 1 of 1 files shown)


GeoffreyDlvl:
This paragraph left me confused when I read it. The documentation actually recommends that props should be named `onSomething` so I think that the `Button` prop should actually be replaced by `onClick` instead of `handleClick`.
"The event handler is passed to the <i>Button</i> component through the _handleClick_ prop." would indeed be the correct sentence with the current code but then "React's own official [tutorial](https://reactjs.org/tutorial/tutorial.html) suggests this convention." would be a wrong statement.
I wanted to publish a PR that would replace the prop name with `onClick` but in part1d there's a screenshot with the prop name, and I couldn't reproduce the exact same visual to keep it consistent with other screenshots.

PR #2728, opened 2023-03-13
Typo

Benbb96:


dylancdavis:
Is "Cours" not an acceptable translation of "Course"?

PR #2735, opened 2023-03-14
Update part5d.md

sahandsn:
cross-env is needed, for windows at least


mluukkai:
there will be a rewrite in the part

PR #2751, opened 2023-03-19
Update part7b.md

cabbachew:
Removed link to [old React Router page on hooks](https://reacttraining.com/react-router/web/api/Hooks) because the content no longer exists. The [new docs](https://reactrouter.com/en/main) do not have a route to an overview of hooks because documentation for each hook is directly accessible via the sidebar. Since the link to the new docs are already embedded earlier in the sentence, this link is no longer necessary.


cabbachew:
In the same section ("Hooks") in Part 7b, there are links to pages in the old React docs that all have a counterpart in [the new React docs](https://react.dev/):

image

I did not create a pull request to change them because the old docs are still available despite not being updated. I considered raising an issue, but if it would be helpful to update these links I can propose those changes as well.

PR #2771, opened 2023-03-25
Update part10c.md

AhmedKasu:
I believe we should import the authStorageContext that we already created in the AuthStorageContext.js file instead of creating another context in the useAuthStorage.js file.  The authStorageContext from AuthStorageContext is the one we are already using as the context provider in app.js


nikogenix (approved):
yep, I just ran into an issue because of this as well. in its current state, the hook leads to an undefined `authStorage` when used as instructed a bit later in this lesson.

```js
import { useAuthStorage } from '../hooks/useAuthStorage';
const authStorage = useAuthStorage();
await authStorage.setAccessToken(/* access token from the data */); // undefined
```

making the same changes as the PR, everything works as expected.

PR #2785, opened 2023-03-28
Exercise2.18-2.20_changes

oodtoon:
Feedback on exercises 2.18-2.20.


akirataguchi115 (changes requested):
Could you rather make an issue of this problem and close this pr or maybe just write the actual changes you want to see to the html?

mluukkai:
indeed, this should be an issue

PR #2798, opened 2023-04-01
Part 9b: fix bug in parseArguments function

JVPH:
Fix bug where passing an empty string '' as an argument would not throw an error.


mluukkai:
you are correct yes, but I will not merge this since being prepared for an empty string doesn't not make much sense in a toy example like this.

PR #2802, opened 2023-04-01
Update part3c.md

MarcGuay:
Explicitly state that we're going to be switching to working on the notes app backend


mluukkai:
well, it does not harm to add these but in parts 3- there is just a single app that is worked with in exercises and that should be pretty obvious from the context

PR #2819, opened 2023-04-06
Update part4a.md

MarcGuay:
Not sure if this is a mistake or if we are expected to put the `return 1` in there ourselves


mluukkai:
you have to put it yourself

PR #2829, opened 2023-04-08
Update part3d.md

chmromano:
Removed validation ambiguity from exercise 3.20, fixed typo.


chmromano:
It would also be valid for the text to be in the form:

- if formed of two parts that are separated by -, the first part has two or three numbers and the second part also consists of numbers
    - eg. 09-1234556 and 040-22334455 are valid phone numbers
    - eg. ~~1234556, ~~1-22334455 and 10-22-334455 are invalid

Right now it is unclear whether the phone numbers are only valid if inputted with a -, or if both phone numbers with - and phone numbers with only numbers are.

PR #2842, opened 2023-04-11
Update part7e.md

sahandsn:
simple grammar error


akirataguchi115 (commented):
every platform != as many platforms as possible; if this was the intention then LGTM!

PR #2852, opened 2023-04-13
Update part6c: createNote action in noteReducer.js

pongbao:
Generation of id is not included in the addNote function in NewNote.js and so we need to add the id property within noteReducer itself


mluukkai:
backend creates the id

pongbao:
> backend creates the id

Oh right, it totally slipped my mind that json-server does the id generation. Thanks!

PR #2863, opened 2023-04-20
Part4d: token existence check is not reached

marcopdonoso:
Apparently, this verification cannot be achieved, because if there is no token the jwt.verify() method will throw a JsonWebTokenError exception if token is equal to null.


vejol:

PR #2867, opened 2023-04-25
Update part8b.md

MyatKaungKyaw:
there is no error property in extensions. So I change the code as above.


vejol:
Thank you for the pull request! Error handling has changed due to the Apollo Client version update. The material has now been updated, so this fix is no longer needed.

PR #2874, opened 2023-04-28
Added clarification on requirements for exercise 3.4 and 3.5

IggsGrey:
Clarified whether changes to the phonebook are to be persisted over time or not.

Please help with links if the attached ones are inappropriate

- [Reading files in Node.js](https://www.geeksforgeeks.org/node-js-fs-readfilesync-method)
- [Writing files in Node.js](https://www.geeksforgeeks.org/node-js-fs-writefilesync-method)


mluukkai:
oh no, this is not needed for these exercises

PR #2879, opened 2023-04-29
Fix typo in part9c.md

fn-Drew:
Fix typo in the sentence "VS code looses track ", so the wrongly typed word "looses" is replaced with the word "loses" as expected.


akirataguchi115 (approved):
LGTM!

PR #2889, opened 2023-05-08
update page for zh 2023

RichardStark:


RichardStark:
[WIP ALERT!]I translated this course again with chatGPT, some content needs to be reviewed, so the status is WIP.

PR #2894, opened 2023-05-12
Update part4c.md

chriscodes17:
In the creating a new note section, when saving the id of a user to a new note, it states that the note scheme will also need to change - {user: String}.  But in the sections of 'Mongoose schema for users' and at the end of populate, the note schema shown is different from the one stated above. Was there a reason to state them differently. I am seeing it as conflicting information with both being different. This is a proposal to remove the {user: String} and use the Object.id type.


mluukkai:
taken care in another commit

PR #2897, opened 2023-05-15
Update part8c.md

Madobyte:
The `createUser` mutation does not have the `name` property in the `args` parameter.


vejol:
Thank you for the pull request! These have already been fixed in another commit.

PR #2912, opened 2023-05-23
Changed the /api/testing/reset endpoint to accept a DELETE request

vkats90:
Since this endpoint clear the database for the purpose of testing a DELETE request is more appropriate than a POST one


vkats90:
Done, pull request #34 on the example code

mluukkai:
well, decided that this won't be changed. despite the endpoint being used to delete stuff, this is a bit special endpoint and and POST is semantically better

PR #2923, opened 2023-05-31
Update part1a.md

nrsmac:
below code resides in index.js


lorenzojkrl (changes requested):
Thanks for your comment.
The React component named App is correctly defined in App.js.
The App component is then imported and used in index.js.
I don't think this PR should be merged.

Leagian:
as Lorenzo said, it is imported in index.js indeed

PR #2930, opened 2023-06-06
Update part11b.md - Add back alternative e2e workflow

nikogenix:
I found that this was first brought up in #2204 and it was further improved with commit 42ac1ca, but it was later removed in 23cd3ff.

in my case, after more than an hour of research and troubleshooting, I couldn't find any other solution besides this old fix.

if someone has a better idea, I'd be glad to test it out on my setup, since this doesn't seem to be a universal problem.

here's the workflow in my repo for context:
https://github.com/nikogenix/fso-part11/actions/workflows/pipeline.yml


mluukkai:
cypress no more used

PR #2932, opened 2023-06-07
style(thumbnails): add color to thumbnails

mayconblopes:
# Colorful thumbnails

https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/assets/614145/8bac4db8-1d96-407a-bdbf-ef5ee15ea48b


vejol:
Thank you for the pull request, we greatly appreciate the work you've done!

The proposed changes seem a bit problematic to me, though. Adding colors to the thumbnails makes the front page look cluttered, so I think it's better for the thumbnails to be black and white by default and colored when hovered over.

Another issue is that similar images are used as colored background images for the parts, and the thumbnails you proposed do not match the coloring of the part background images:

image


Unfortunately, we have very little time to work on pull requests, and it seems this one has already been open for quite a while. I will close it now, but if you want, feel free to revisit this pull request and develop it further!

PR #2938, opened 2023-06-10
Update part1d.md

eslamalbaik:
The component gets access to the functions _setLeft_ and _setRight_ that it can use to update the two pieces of state.


lorenzojkrl (changes requested):
This is only repeating the previous sentence.
Is there anything I am missing?

paulotijero:
@lorenzojkrl Yes, you're right, it doesn't seem like a change.

Leagian:
yep it is the same sentence

PR #2944, opened 2023-06-14
Removed redundant  release_command from fly.toml

laurimaila:
As stated [here](https://fly.io/docs/reference/configuration/#run-one-off-commands-before-releasing-a-deployment), the release_command runs in a temporary VM and _Changes made to the filesystem on the temporary VM will not be retained or deployed. The building/compiling of your project should be done in your Dockerfile._ 
This also means that webpack doesn't need to be moved to Dependencies since the building is done in the Dockerfile with devDependencies.
Tested that this works in my own fork.


mluukkai:
fixed in another commit

PR #2952, opened 2023-06-19
Update part10c.md

D3lux3:
Added instructions on how to get pass errors while installing dependecies.


mluukkai (changes requested):
the course is using npm

PR #2955, opened 2023-06-23
Update part0a.md

moonergone0:


src/content/0/en/part0a.md
```diff
@@ -1,5 +1,5 @@
 ---
-mainImage: ../../../images/part-0.svg
+main image: ../../../images/part-0.svg
```
mluukkai, line 2: this can not be changed

PR #2963, opened 2023-06-24
Update part3a.md (Math.max explanation moved to top)

aju-alen:
Problem Statement 
When we first see the piece of code ``` Math.max(...notes.map(n => n.id)) ```, it is a bit intriguing and I had to look for other resources to see how it works.  The explanation to the line of code was at the end of the notes.

Solution 

To move the explanation right after the piece of code was first introduced.


mluukkai:
will change in another commit

PR #2977, opened 2023-06-27
Update part11a.md

hobi9:
fix typo


mluukkai:
this is intentional, see https://www.dictionary.com/browse/automagically

PR #2993, opened 2023-07-04
Update part4a.md

kristianka:
Add English and Finnish instructions for ```jest``` testing if user is using ```import``` statements instead of ```require```


mluukkai:
this content is removed from the material

PR #3000, opened 2023-07-07
Update part1d.md

Nathaniel330:
removed negative sign in: the average score (good: 1, neutral: 0, bad: -1)


Nathaniel330:
closed. i thought it was a typo. turns out it is used to calculate the average score. :)

PR #3016, opened 2023-07-12
Add [env] if [http_service] configuration

SpencerWightman:
Otherwise defaults to PORT 3001:
const PORT = process.env.PORT || 3001


SpencerWightman:
The issue (fly.toml adjustment) was resolved independently of this ticket.

PR #3019, opened 2023-07-12
Remove redundant "fly deploy" in part3b

AlecBlance:
```fly deploy``` was mentioned 2 times which might confuse readers


mluukkai:
this is not redundant

PR #3021, opened 2023-07-14
Update part4d.md - 'Bearer' / 'bearer' case mismatch

SpencerWightman:


SpencerWightman:
Consistency: 'bearer' is used in the following 'REST client' and 'Postman' images.

mluukkai:
Bearer is the right one, the images should be fixed

PR #3038, opened 2023-08-02
Update part1c.md

iamaamunir:
Calling ReactDOM.createRoot() inside the refresh function returns an error in React 18.


mluukkai:
this can not be right

PR #3049, opened 2023-08-08
Update part4c.md

jeanmadao:
Delete the part where we had to switch the user's type from 'mongoose.Schema.Types.ObjectId' to 'string' for no good reason


mluukkai:
deleted by another commit

PR #3072, opened 2023-08-17
Update part13b.md - fix table formatting

tylersernett:
Add spaces to fix SQL table format


mluukkai:

PR #3081, opened 2023-08-24
Use vite to initialize new application in part Exercise1.1-1.2

rKhanna15:
main.jsx and app.jsx can be accessed if vite is used to initialize the react app and not create-react-app


mluukkai:
thanks!

PR #3083, opened 2023-08-25
update part4c.md

makzuu:
The toJSON method has not been called yet, so the id property doesn't exist. (I tested it locally, but maybe I'm wrong)


mluukkai:
you must be wrong...

PR #3094, opened 2023-08-31
Update part0b.md

akankha:
typo


akankha:
just correct the typo

PR #3113, opened 2023-09-10
Update osa6e.md

Wadezkii:
fixed broken link


Wadezkii:
old link was missing a letter and thus was invalid

PR #3137, opened 2023-09-19
Update part3b.md

devpius:
The vite configuration for the proxy didn't work, i still get a 404 on /api/persons untill i added a rewrite regex key to the object to replace the any matches with path /api for any requesr

### DIFF
--- a/src/content/3/en/part3b.md
+++ b/src/content/3/en/part3b.md
@@ -474,6 +474,7 @@ export default defineConfig({
       '/api': {
         target: 'http://localhost:3001',
         changeOrigin: true,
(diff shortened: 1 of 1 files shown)


Alkebsi:
To me, I haven't even been able to use Vite. I manually configured the proxy after spending more than 30 min. I think they should have mentioned how will someone not using vite deal with the such a situation! They will save a lot of their students' time.

pabloaugustobarragan:
Heroku is not free, it is paid

mluukkai:
this problem seems not to exist anymore

PR #3138, opened 2023-09-20
Update part3c.md

SlinkyPotato:
fix id key name


mluukkai:
frontend assumes the field id (not _id)

PR #3141, opened 2023-09-22
Update part1a.md, there is a small typo error in the command

kutt27:
Changed the code from 

```
npm create vite@latest part1 -- --template react
```

to 

```
npm create vite@latest part1 --template react
```


SlinkyPotato (commented):
This is intentional. The "--" is used to seperate npm arguments from the vite arguments. The latter args "--template react" are passed to the vite cli 

https://stackoverflow.com/questions/43046885/what-does-do-when-running-an-npm-command

PR #3142, opened 2023-09-22
Part5e: Resolve testcase failing.

YogeshDollin:
Resolve failing test case issue.
if we skip the "show all" click step, line 540 fails because after making a note "not important", it filters out that note due to "show all" state.


vejol:
Thank you for the pull request. The Note App has now been changed so that the "show all" mode is enabled by default. Therefore, this change is no longer needed.

PR #3147, opened 2023-09-26
Corrected English in 3b

popovidis:
`Fly.io might require one to enter the credit card details` sounds awkward, `Fly.io might require one to enter their credit card details` sounds correct.


popovidis:
I created a new pull request after noticing the same occurrence again.

PR #3169, opened 2023-10-07
Update part0a.md

alecjoseph-dev:


alecjoseph-dev:
Edited this part:

> In parts 1-4 you are expected to do `at least `all of the exercises that are not marked with an asterisk(*). Exercises marked with an Asterisk count towards your final grade, but skipping them does not prevent you from doing the compulsory exercises in the next parts. Parts 5-13 do not have Asterisk marked exercises since there is no similar dependency on previous parts.

removed at least and made `all` in bold

PR #3185, opened 2023-10-15
Update part12a.md

aguperes:
Link to the official image of node on docker hub is broken. Replaced it for the one that works


lorenzojkrl:
Good catch, and now I learned URLs are case sensitive.

PR #3188, opened 2023-10-15
Update part6d.md

Pulkitxm:
I have a small correction in the course material [here](https://fullstackopen.com/en/part6/react_query_use_reducer_and_the_context#managing-data-on-the-server-with-the-react-query-library:~:text=import%20%7B%20useQuery%20%7D%20from%20%27react,in%20the%20branch%20part6%2D1.) in the [part 6(d)](https://fullstackopen.com/en/part6/react_query_use_reducer_and_the_context) 
image

here the `querFn : getNotes` is generating an error as **_notes.map is not a function_** 

image

which is getting resolved if I include parenthesis with it just like this `querFn : getNotes()`

image

So I would request you to check this out

image

### DIFF
--- a/src/content/6/en/part6d.md
+++ b/src/content/6/en/part6d.md
@@ -149,7 +149,7 @@ const App = () => {
 
   const result = useQuery({
     queryKey: ['notes'],
-    queryFn: getNotes // highlight-line
+    queryFn: getNotes() // highlight-line
   })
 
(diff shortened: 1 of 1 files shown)


mluukkai:
it is right currently, there is an error in your 2nd pic where you have 

queryFn: () => getNotes

PR #3196, opened 2023-10-19
Update part3c.md (ZH)

peisong13:
In Part 3c (ZH), sections below 'Error Handling' is not translated in to Chinese. As a Chinese native speaker, I would love to contribute.


mluukkai:
Good job!

PR #3199, opened 2023-10-19
Update part6d.md

Pulkitxm:
I have a small correction in the course material [here](https://fullstackopen.com/en/part6/react_query_use_reducer_and_the_context#managing-data-on-the-server-with-the-react-query-library:~:text=import%20%7B%20useQuery%20%7D%20from%20%27react,in%20the%20branch%20part6%2D1.) in the [part 6(d)](https://fullstackopen.com/en/part6/react_query_use_reducer_and_the_context) 
image

here the `querFn : getNotes` is generating an error as **_notes.map is not a function_** 

image

which is getting resolved if I include a function call with arrow func() with it just like this `querFn : ()=>getNotes`. So I would request you to check this out and I made the changes you suggested  #3188 

image

### DIFF
--- a/src/content/6/en/part6d.md
+++ b/src/content/6/en/part6d.md
@@ -149,7 +149,7 @@ const App = () => {
 
   const result = useQuery({
     queryKey: ['notes'],
-    queryFn: getNotes // highlight-line
+    queryFn: ()=>getNotes // highlight-line
   })
 
   // ...


kahlstrm:
I can't reproduce this, so seems like there might be some error in your code?
Can you try following the following steps:
1. Clone the example application [query-notes](https://github.com/fullstack-hy2020/query-notes/tree/part6-1)
2. Switch to the example branch `git switch part6-1` 
3. Run `npm install` and `npm run dev`
4. On another terminal, run `npm run server`
5. navigate to the frontend (http://localhost:5173) and see if this happens.
Following these steps I couldn't reproduce this.

PR #3205, opened 2023-10-21
Update part1d.md - missing return() in Button component

inshal-haq:
in part1d, at the end of Conditional rendering, the Button component needs to have return () to render button

p.s. thank you for these valuable lessons fso!

### DIFF
--- a/src/content/1/en/part1d.md
+++ b/src/content/1/en/part1d.md
@@ -414,9 +414,11 @@ const History = (props) => {
 
 // highlight-start
 const Button = ({ handleClick, text }) => (
-  <button onClick={handleClick}>
-    {text}
-  </button>
+  return (
+    <button onClick={handleClick}>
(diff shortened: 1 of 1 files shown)


inshal-haq:
Also can update in "Debugging React applications" section, when Button component is re-shown

kahlstrm:
The syntax in the example is correct, arrow functions do implicit returns with the `() => expression`-syntax, see [Arrow function expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#description).

inshal-haq:
Oh wait, I just noticed this. My mistake, I was using curly braces rather than the parantheses that they are using

Thank you! closing...

inshal-haq:
Incorrect commit, code is working as intended, using implicit return in arrow function

PR #3210, opened 2023-10-23
Update osa3c.md

Nerdman4U:


vejol:
There is no error in the original command. The application is still intended to be started with Node but using [debug mode](https://nodejs.org/en/learn/getting-started/debugging).

PR #3220, opened 2023-10-26
Updated instructions for installing Node

ancxanas:
Current instruction will be outdated in some time.


ancxanas:
Removed sudo from the commands.

PR #3231, opened 2023-10-30
Update part7d.md - replace semicolon with comma

bambery:
The build will fail if the semicolon is placed here instead of a comma. Replacing with a comma allows the build to succeed.


bambery:
Additional semicolons that don't affect the build were also deleted somehow. I will readd them and resubmit

PR #3236, opened 2023-10-31
Update part6d.md

bambery:
Code as written will throw a `defaultMutationOptions is not a function` error. Changing the code to include a `mutationFn` key will solve the problem.


bambery:
didn't see PR #3214 which solves the same issue

PR #3245, opened 2023-11-03
Update part1a.md

RexRowan:
Deleting the Hello component would cause the page not to load and errors in the DOM.


mluukkai:
well, it should be obvious that you can not delete a component. and if you delete it, there will be an important lesson. so will not add this. would be too much hand holding

PR #3246, opened 2023-11-05
Fix button onClick typo in part1c.md en/fr

willylin207:
The `<button>` react component needs to use onClick instead of onSmash for the event handler property.


KartikAroraOfficial:
Hey, No.
the onSmash there was justified and was added there according to the react official docs, which have been cited there. if you need further info. you can refer to this following article from the react's official documentation. 
[Naming event handler props](https://react.dev/learn/responding-to-events#naming-event-handler-props). 
Thus, your PR should not be merged into the source repo.

PR #3248, opened 2023-11-05
Update part3b.md note about "dist" in .dockerignore

SebastianEShelby:
Updated the note section about fly `.dockerignore` file to be more clear and in accordance with the course material (Note that the front-end project creates a "dist" folder not a "build" folder).


mluukkai:
fixd by another commit

PR #3250, opened 2023-11-06
change onClick for onSmash in the last line

ruperlopez:
last line is
const Button = ({ onSmash, text }) => <button onSmash={onSmash}>{text}</button>

it should be:
const Button = ({ onSmash, text }) => <button onClick={onSmash}>{text}</button>

since it is the browser button.
this can be found in: https://react.dev/learn/responding-to-events#naming-event-handler-props


KartikAroraOfficial:
yes, you're right. but another [Pull Request](https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/pull/3246#issuecomment-1794555431) has already resolved it for both english and french parts.

PR #3251, opened 2023-11-06
Update part4b.md

ShivangBhawsar:
I got this error ("TypeError: Note.findByIdAndRemove is not a function," ) while deleteing a note with mongoose v 8. I think this is because the "findByIdAndRemove" function has been deprecated and "findByIdAndDelete" is the alternative.


mluukkai:
changed in another commit

PR #3254, opened 2023-11-07
Update part3c.md - MongoDB Model methods

inshal-haq:
In part3c, section "Other operations", the correct method for delete operation is findByIdAndDelete(). Possibly MongoDB updated it from Remove to Delete.

Also updated the links to properly open the method definitions


mluukkai:
changed in another commit

PR #3267, opened 2023-11-11
Update part4a.md

inshal-haq:
In section "Note on exports", the final point "Finding the usages of your exports with VScode" has some visual issues and is overlapping the sidebar.

The only fix I committed is grammar corrections for this final point. But still needs the visual correction.


inshal-haq:
Attached screenshot of visual issue. I am using chrome on macOS
Screenshot 2023-11-11 at 3 47 06 PM

PR #3279, opened 2023-11-16
Update part4c.md

enzolonghi:
The version 6.12.2 of Mongoose is also not compatible with the latest version of mongoose-unique-validator. You need to install version 7.6.5 of Mongoose for it to be compatible.


mluukkai:
fixed with another commit

PR #3291, opened 2023-11-22
Add "Hide extension URLS" recommendation

smith558:
Add "Hide extension URLS" recommendation. Otherwise the requests of any extensions (which we all have nowadays) clutter the "Networks tab".


mluukkai:
this is a good addition!

PR #3292, opened 2023-11-22
Update part5b.md

myverdict:
Added `export default NoteForm` to NoteForm component.


mluukkai:
these are quite often intentionally left out since reader is supposed to know at this point that there needs to be a export

PR #3294, opened 2023-11-23
Add missing `const` in delete route declaration

jeremytsang1:
For consistency with the rest of the text and also there seems no need
to create a property on the global object at this point.


jeremytsang1:
My mistake, forgot about the global notes object.

PR #3295, opened 2023-11-24
Update osa1d.md

areee:
Add the missing function definition. Before this fix, this exception was thrown to the console:
> Uncaught Error: Too many re-renders. React limits the number of renders to prevent an infinite loop.


vejol:
Fixed in other commit.

PR #3326, opened 2023-12-07
Update part1d.md

avdiutaulant:
In the section (Do Not Define Components Within Components) I make some changes in code because you forgot to make return statement in Display, Button and after these changes the code works perfectly because of that code that you forgot the code doesn't show anything after these changes I hope other students will be fine during writing the code


mluukkai:
return is not needed when function has just one statement

PR #3328, opened 2023-12-08
Add a note to lesson part1-c

rosa-pastel:
Students are told that prop name `onClick` can be changed to `onSmash` but this change on its own creates an error because the prop name passed from App to Button doesn't match the one used inside Button. This note can be added to avoid confusion.


vejol:
Fixed in other commit.

PR #3339, opened 2023-12-18
Update part3b.md

Yisaaaa:
replaced comma with em dash to make the text more easier to understand


mluukkai:
not sure if this makes it better

PR #3348, opened 2023-12-24
Fixed the phrasing of the Ninety–ninety rule

yoniLavi:
Note - https://en.wikipedia.org/wiki/Ninety%E2%80%93ninety_rule

Here's the original Programming Peals article where Jon Bentley attributed this aphorism to Tom Cargill from Bell Labs: https://dl.acm.org/doi/pdf/10.1145/4284.315122

### DIFF
--- a/src/content/1/en/part1d.md
+++ b/src/content/1/en/part1d.md
@@ -1291,7 +1291,7 @@ const App = () => {
   const anecdotes = [
     'If it hurts, do it more often.',
     'Adding manpower to a late software project makes it later!',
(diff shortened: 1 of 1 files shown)


yoniLavi:
Looking at the file's history, I just noticed it was correct at the start and mis-fixed in this commit a couple of years ago:
https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/commit/c6a7b70626b955d066ebc37af2a2d93a707c0b78

PR #3389, opened 2024-01-13
Update part3c.md

gitairman:
When I tried to use the original code I kept getting errors thrown.  When I changed it to run in an async function as per the Mongoose documentation, it worked ok.


mluukkai:
mix and match of async and then is not a good idea...

PR #3390, opened 2024-01-14
Update incorrect example code

rojasjosiah:
Simple change, but the changed segment shows what the current code looks like. Not only is it missing a portion of what it actually looks like, but without it, nothing shows.


mluukkai:
the code is valid

rojasjosiah:
These components need to be contained within the return statement. If you run the code without that, nothing is returned and the component is not displayed.

PR #3398, opened 2024-01-20
Notes "user" field changed to "userId"

David4bay:
I was getting an error when I was creating new notes referencing the user id and tracked it down to the note schema not referencing the id properly, changing the "user" field to "userId" fixed it.


mluukkai:
this is not right

PR #3415, opened 2024-01-26
part 3b - `fly open` to `fly apps open`

adlantz:
When `fly open` is run fly gives the following response:

```
Command "open" is deprecated, use `fly apps open` instead
```


mluukkai:
fixed in another commit

PR #3416, opened 2024-01-27
Added labels to input fields

DrMoshtael:
It's better for accessibility to associate inputs with a label. Also, I don't think the "name" attribute was doing anything as we're using state, so it's confusing to have it in there.


vejol:
Thank you for the pull request! Label elements have now been added to the inputs in the material. I made this change in separate commits, as it required modifications in many places throughout the material.

PR #3438, opened 2024-02-07
Update osa0b.md - Replace marketing website URL with MDN URL

JuhQ:
httpstatuses.com seems be owned by some marketing company these days and I don't think this URL should be used. MDN is a better resource for this.


JuhQ:
HOX: I've not checked if the same domain is used elsewhere in the material.

PR #3452, opened 2024-02-12
Fix sentence

micpob:


mluukkai:
not a typo, see https://dictionary.cambridge.org/dictionary/english/clean-slate

PR #3471, opened 2024-02-23
Update osa5b.md

justenlu:
It seems to me that the whole blog is NOT replaced with update data, but only those fields that have defined value are updated. Could you please check this?


vejol:
Thank you for the pull request! The purpose of the PUT method is to send the entire updated object, not just a specific field to be updated. It is possible that the backend is implemented in a way that sending only the `like` attribute is sufficient, but that is not in line with conventions. 

The functionality you described is typically associated with the http PATCH method, which is not used in this course. Therefore, I will not accept this pull request.

PR #3472, opened 2024-02-23
Update part4b.md - move after inside `describe`

anantakrroy:
`after` hook runs correctly only inside a test suite. Using it as standalone outside the suite i.e outside the `describe` block causes the test to hang indefinitely .


mluukkai:
what node version you are using? should work with 20. the following now added  to material

_**Note** this course material was written with version v20.11.0 of Node.js. Please make sure that your version of Node is at least as new as the version used in the material (you can check the version by running node -v in the command line)._

PR #3480, opened 2024-02-27
Update part4d.md

gurramnikhil4:
User ids are referred through user.id and not through user._id since we've deleted _id through the transform option in the note model


mluukkai:
it is only the json representation where the _id has been removed. this user comes from db

PR #3482, opened 2024-02-29
Update part2c.md

sanberkzulfikar:
As stated JSON-Server CLI when run: "--watch/-w can be omitted, JSON Server 1+ watches for file changes by default"

--watch flag is not needed anymore.


sanberkzulfikar:
In the npm section there's also an unnecessary use of --watch flag.

PR #3484, opened 2024-02-29
Corrected filename from diaries.ts to entries.ts

DrMoshtael:
The example code has the file that was previously entries.json become entries.ts. However, the description calls it diaries.ts. It is imported as entries.ts below.


mluukkai:
fixed in new material

PR #3489, opened 2024-03-02
Update part12c.md

lxhelmer:
Using the general filename here instead of the explicit development environment filename as in the displayed tree is confusing since the general filenames match those used in the exercise 12.20 production environment.


mluukkai:
will soon be fixed with another commit

PR #3490, opened 2024-03-02
Update part12c.md

lxhelmer:
The filename doesn't match the naming stated prior in the same section. This causes unnecessary confusion, since the same filename is used in exercise 12.20 in the production environment.


mluukkai:
will soon be fixed with another commit

PR #3499, opened 2024-03-08
Update part7b.md

Yisaaaa:
The data returned from the API is of different structure from what is written in the starter code. Saying that the code is almost complete aside from the to be implemented useEffect code is a bit misleading.


mluukkai:
exercise removed

PR #3500, opened 2024-03-08
Part 10c: Environment variables now supported

DrMoshtael:
"Unfortunately, React Native doesn't have direct support for environment variables." Is this now false? I don't know about React Native, but it seems Expo SDK 49 added support:  https://docs.expo.dev/guides/environment-variables/


mluukkai:
fixed in new material

PR #3504, opened 2024-03-10
Update part2c.md

kaipoarita:
Added parentheses to hook() makes it a function call,  resolving the issue of data not loading on the first page render.


vejol:
The first parameter of useEffect should be a function, not the result of a function call. Therefore, parentheses are not needed.

PR #3506, opened 2024-03-12
Fix grammar in part0a.md

aleichtm:


aleichtm:
other options could be instead changing the word "despite" to "but", "even though", or some other option

PR #3522, opened 2024-03-15
Update part11b.md

LerkkaP:
Remove 'command: npm run test:e2e' from the last exercise workflow. This command causes the workflow to stall because it starts Cypress in interactive mode, requiring manual interaction.


mluukkai:
cypress no more used

PR #3524, opened 2024-03-17
Update part1a.md

santylaborde:
File name is wrong


santylaborde:
I was wrong, the component was defined in App.jsx and the main.jsx was rendering its content.

PR #3535, opened 2024-03-22
Removal of "Taken into use" and other phrasing changes.

stritch-dev:
'Taken into use' is an awkward phrase. Replaced it with more natural (and hopefully clearer) wording. Made some other phrasing changes.


stritch-dev:
Commit b2c9c01 removes the term "so-called" since it implies an inaccurate  name. https://dictionary.cambridge.org/us/dictionary/english/so-called

PR #3559, opened 2024-04-04
Whitelist instructions

yyliang211:
Add note that you need to whitelist all IP addresses in MongoDB Atlas for the fly.io app to work


yyliang211:
seems like a common problem
https://community.fly.io/t/shared-ipv4-and-whitelisting-woes/12545

PR #3561, opened 2024-04-04
Add import for 'assert'

ash-cache:
The code example for grouping tests with `describe` uses `node:assert`, but the line to import it is missing. Without it, there is an error when running the test.


EledenGreen (commented):
i think it is to be understood under: `// ...`
But yeah. would make the instruction more clearer if the statement is included.

vejol:
Fixed in another commit.

PR #3570, opened 2024-04-09
Add command to start the updated backend and edit the database URL

A7YosSdHF5fb4:
1. Add the command `node index.js yourpassword` to describe how to start the updated backend. Although students are supposed to be able to infer this, omitting it may cause confusion for some, as the previous command used was `node mongo.js` yourpassword.

2. Edit the `url` link to include the new name `noteApp` for the database, as this is the database we were working with in the `mongo.js` file.


vejol:
Thank you for the pull request! This was fixed in another commit because of conflicts.

PR #3571, opened 2024-04-09
fix several mispellings in part 12c

jlyon12:
corrects three instances of 'othervise' to 'otherwise'


EledenGreen (commented):
typo has been fixed

PR #3579, opened 2024-04-12
Fix: inconsistencies in ports, improved wording

laemoi:
There seems to be some inconsistencies with regard to the ports that are being used. According to the images in the material (attached below), the frontend is running on port 3000, while the backend is running on port 3001. However, in the text, the frontend is always referred to as localhost:5173.

image


mluukkai:
the figure has the wrong port (it is from time create react app was used and there 3000 was the default)

PR #3582, opened 2024-04-16
Käynnistäessa sovelluksen selain ilmoittaa "sivustoon ei saada yhteyttä"

TainaLepisto:
https://stackoverflow.com/questions/70694187/vite-server-is-running-but-not-working-on-localhost

https://github.com/microsoft/vscode-remote-release/issues/8213


vejol:
This is a special case and a rather rare error, so we have chosen not to include it in the material.

PR #3599, opened 2024-04-26
Update part4c.md

respp:
The small sentence that they forgot to translate into Spanish on line 217


vejol:

PR #3600, opened 2024-04-27
fix vite configure proxy

RuilinYang-beta:
Added a line to correctly configure proxy server of the frontend App. 
See Vite docs: https://vitejs.dev/config/server-options.html#server-proxy


vejol:
Thank you for the pull request. There is no need to rewrite the path here because the backend is configured to listen to routes /api/notes and /api/notes/1 etc. We don't want to remove /api part from url.

PR #3601, opened 2024-04-27
fix(test): `.forEach` expects a synchronous function

ricky-kiva:
Some tests related to database operation over API could fail erratically. In my case, `a specific note can be viewed` test failed. This happened because `.forEach` expects a synchronous function. 

Reference: https://betterprogramming.pub/javascript-async-await-pitfalls-with-foreach-d55dddc42d0


ricky-kiva:
Turns out, it's already explained that `.forEach` won't work within `beforeEach()` in the later part.

PR #3606, opened 2024-04-30
Update link to Mozilla JavaScript Overview on part1b.md

aleichtm:
old link redirects to this page that contains the same content with a different title and url


EledenGreen (approved):
The link does redirect to **JavaScript language overview**

PR #3608, opened 2024-05-01
Part5d: Fix description in exercise 5.20 with Playwright E2E

jv-soares:
I think that the description in exercise 5.20 with Playwright E2E is currently wrong.

The description tells us to make a sure a blog can be edited, although this functionality was not required in previous exercises. In addition, the description for the corresponding exercise in the Cypress E2E chapter tells us to ensure users can **like** a blog.


vejol:
Thank you for the pull request! This has been fixed in another commit

PR #3632, opened 2024-05-18
Update part2.mb 2.7 link

Presssure:
The link in the exercise 2.7: The Phonebook Step 2 is incorrect and has been changed to: https://www.joshbritz.co/blog/why-its-so-hard-to-check-object-equality/


EledenGreen (approved):
updated link works

PR #3636, opened 2024-05-20
Update part2c.md

lvu5:
I made change to make sure that id's datatype is string to access individual in the server.


lvu5:
Review!

mluukkai:
changed by other commit

PR #3639, opened 2024-05-27
Update osa13c.md

justenlu:
Shouldn't we use camel case here?


mluukkai:
fixed in another commit

PR #3645, opened 2024-05-29
Update part3d.md

smwingira:
Updated the ESLint instructions to correspond with the latest flat configuration format as per the [ESLint guide](https://eslint.org/docs/latest/use/getting-started).


EledenGreen (commented):
eslint is a mess for me 😃 

i just use : `npm init @eslint/config@latest` 
and Prettier to format: `npm install eslint-config-prettier`

That's it for me.  ✨

mluukkai:
fixed by another commit

PR #3650, opened 2024-05-31
Update part5c.md

myverdict:
The code does not work without { exact: false } for the `findByText` method. So, I added it and the test passes.


EledenGreen (commented):
it should work without { exact: false } if we are using findByText

PR #3652, opened 2024-06-02
docs: replaced a dead link with a similar link from MDN docs

madhusudan-kulkarni:


EledenGreen (approved):
The MDN doc is much easier to understand than the original ECMAScript 6

PR #3659, opened 2024-06-04
Update part5d.md

imandresmorales:
[en] I got this error frequently in exercise 5.20. Although it may appear in others, I wasn't sure where to add it, whether before starting the exercises or in exercise 5.20 which was where I had that error. [es] Recibí este error con frecuencia en el ejercicio 5.20. Aunque puede aparecer en otros, no estaba seguro en dónde agregarlo, si antes de comenzar los ejercicios o en el ejercicio 5.20 que fue donde tuve ese error.

### DIFF
--- a/src/content/5/es/part5d.md
+++ b/src/content/5/es/part5d.md
@@ -1299,7 +1299,7 @@ La prueba debe garantizar que un nuevo blog es visible en la lista de todos los 
 
 #### 5.20: Pruebas de Extremo a Extremo de la Lista de Blogs, paso 4
 
-Haz una prueba que compruebe que el blog puede editarse.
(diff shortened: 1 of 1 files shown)


imandresmorales:
update part 5

EledenGreen (commented):
As mentioned in the material:
"E2E tests can also be [flaky](https://hackernoon.com/flaky-tests-a-war-that-never-ends-9aa32fdef359). Some tests might pass one time and fail another, even if the code does not change at all."
I too encountered a lot of test timeout either because my internet connection was **slow** (_had to switch to a better internet than it started working_) or my test code was **wrong**.
If ur connection is fast enough but ur still encountering timeout, consider using UI mode/ Trace Viewer to verify for any possible error.
Specifically for 5.20, make sure to capture the updated value after the like button is clicked.
This can be accomplished for instance by using a _page.waitForFunction_.

Hope it helped.

vejol:
A 3000 ms timeout should be sufficient for all the exercises in this series. There is probably something else wrong with the test you wrote. Therefore, I won't accept this pull request at this time.

PR #3664, opened 2024-06-08
Fixed missing parantheses

HosenbundIO:
router.post(...) method was missing 1 parenthesis


EledenGreen (changes requested):
The parenthesis should be after the curly braces.

PR #3666, opened 2024-06-08
Update part2a.md

Sanchit33:
added space between step and 10 and removed the "*" from  "2.3*: Course information step 8."


mluukkai:
* should be there

Sanchit33:
okay!

PR #3674, opened 2024-06-12
Update part4d.md

AhmedLebda:
The user returned after db query `await user.findOne({username}` shouldn't return `_id` because we used the json method on user schema to return `id` instead of `_id`, so all lines that used `user._id` are updated to `user.id`


AhmedLebda:
`await user.findOne({username}` returns a query object which in fact contains `_id` it will only return `id` if we invoked the .toJSON() method on it 

sorry for my mistake.

PR #3680, opened 2024-06-14
Update part5d.md

PaulBoye-py:
- After including the {note.content} in a span, the tests break because the command await page.getByText('first note') now returns a span which does not contain the button. The test did not call the getByText method on the 'second note'


EledenGreen (approved):

PaulBoye-py:
closes #3680

PR #3682, opened 2024-06-18
fix: make password length validation work

jordijs:
The password length validation was not being executed as it was not accessing the corresponding position in the array of process.argv

### DIFF
--- a/src/content/3/en/part3c.md
+++ b/src/content/3/en/part3c.md
@@ -138,7 +138,7 @@ Let's not add any code dealing with Mongo to our backend just yet. Instead, let'
 ```js
 const mongoose = require('mongoose')
 
-if (process.argv.length<3) {
+if (process.argv[2].length<3) {
   console.log('give password as argument')
   process.exit(1)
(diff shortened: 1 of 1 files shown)


EledenGreen (changes requested):
the changes are **not correct** to what we are trying to achieve.

Explanation:
It is a simple argument check, to verify whether we have given password as an argument.
It is **_not_** to check the length of the password.
Here, we are expecting the minimum argument to be of length 3, which could be
`node file.js password`
**NOTE**: if the argument length was less than 3 say:
`node file.js`
then we know that password has not been provided.

In ur code:
`if (process.argv[2].length<3)`
for an argument say:
`node file.js password`
this checks whether the **length of the password** is less than 3, which is not we are tying to achieve.

Hope this clarifies ur doubt.

jordijs:
You are totally right. Sorry, I thought that the goal was validating the length of the password but now I see that it was correctly validating if a password was given. Thank you very much for the explanation.

PR #3683, opened 2024-06-20
Update part4c.md to prevent unexpected test failures

VovaTitarenko:
Tests concerning POSTing new notes failed with an enexpected 500 "Internal Server Error" after adding the update-user functionality to controllers/notes.js router since userId is now required for Mongoose to return the corresponding user object. I added a warning that this the exact reason for test failures so that other students don't spend too much time debugging the code.

P.S. For my tests to pass I also rewrote the beforeEach method since, if the userbase is cleared completely everytime, there will never be a stable userId that can be securely passed inside newNote objects to test the POST functionality.

### DIFF
--- a/src/content/4/en/part4c.md
+++ b/src/content/4/en/part4c.md
@@ -448,6 +448,7 @@ notesRouter.post('/', async (request, response) => {
   response.status(201).json(savedNote)
 })
 ```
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request. I fixed this in another commit with slightly more general text. I also added error handling to the code in case the `userId` is not found.

PR #3685, opened 2024-06-21
Change to a gender neutral pronoun.

Lebussy:
Otherwise it is implied that this rule may not apply to non-male programmers.


mluukkai:
👍

PR #3686, opened 2024-06-24
update in the notes array, change the type of id to string

yogicompute:
changed the type of id to string as in the "Fetching a single resource" section we are using === which will only be true if the value matches along with the type.


EledenGreen (commented):
In the current state, changing
`id: "1"`, 
will solve fetching issue.
But there are some inconsistency in example images.
will make an edit soon.

mluukkai:
changed by another commit

PR #3696, opened 2024-07-01
Update array ids to be strings in part3a.md so code snippets work

zachleach:
Notes array contains *integer* ids when subsequent images and code snippets show/assume they are strings. Therefore, update array to use *string* ids instead of integer ids. 

In section "Simple web server", ../../images/3/2new.png displays ids as strings (very end of the section). 

In section "Deleting resources", code snippet comparing request.params.id !== note.id compares a string with an integer, causing the code example to not work as intended.

In section "Receiving data", (e.g., grep "note.id = String(maxId + 1)", the post request code snippets write ids as strings and not integers.

### DIFF
--- a/src/content/3/en/part3a.md
+++ b/src/content/3/en/part3a.md
@@ -187,17 +187,17 @@ const http = require('http')
 // highlight-start
 let notes = [
   {
-    id: 1,
+    id: "1",
     content: "HTML is easy",
     important: true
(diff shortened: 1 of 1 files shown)


zachleach:
This is a duplicate of #3686

mluukkai:
changed by another commit

PR #3702, opened 2024-07-04
change `note` ids into strings

devserkan:
Our `/api/notes/:id` route assumes that the ids are string and do not convert them with `parseInt` or anything else to match with `request.params.id`.


mluukkai:
changed by another commit

PR #3705, opened 2024-07-07
update: part3a.md (GET /api/notes/:id and DELETE /api/notes/:id Endpoints Not Working)

dev-shahed:
The GET /api/notes/:id and DELETE /api/notes/:id endpoints are not functioning as expected. The endpoints should return or delete the note corresponding to the given ID, but they currently do not. This issue is caused by the req.params.id being a string, while the id in the notes array is a number. As a result, the find and findIndex methods fail to match the note.

Steps to Reproduce:
Start the server using node server.js.
Navigate to http://localhost:3000/api/notes/1 or use a tool like curl or Postman to make a GET request. Observe that the endpoint does not return the expected note. Make a DELETE request to http://localhost:3000/api/notes/1. Observe that the note is not deleted.

Solution:
Convert req.params.id to a number before using it to find or delete the note.

Ref: https://prnt.sc/ohggpLOupB_H

### DIFF
--- a/src/content/3/en/part3a.md
+++ b/src/content/3/en/part3a.md
@@ -503,7 +503,7 @@ Let's make the following change to our code:
 
 ```js
 app.get('/api/notes/:id', (request, response) => {
-  const id = request.params.id
+  const id = Number(request.params.id)
   const note = notes.find(note => note.id === id)
   
   // highlight-start
@@ -530,7 +530,7 @@ Next, let's implement a route for deleting resources. Deletion happens by making
 
 ```js
 app.delete('/api/notes/:id', (request, response) => {
-  const id = request.params.id
+  const id = Number(request.params.id)
   notes = notes.filter(note => note.id !== id)
 
   response.status(204).end()


vejol:
Thank you for the pull request! I am unable to reproduce this error. The `id` is stored as a String in the notes object in this part, so there should be no need to convert it to a Number for comparison. Did you remember to use the "===" operator for comparison? I am closing this pull request for now, but if you want to revisit the issue, please post the code you are using.

dev-shahed:
I've noticed that the issue is not with the code itself but with the documentation. The attached screenshot shows a JSON response where the note id is represented as a string within the note object, see here: https://prnt.sc/UPI-vUc520-3
and here: https://prnt.sc/qSxhc3aRVhqe

 However, in the actual code, note id is a number: https://prnt.sc/bC9sftYfngRC 
Also, within the response we've got: https://prnt.sc/hZBJ5xW-3p2Z

This inconsistency may cause confusion for learners. It would be beneficial to update the documentation to reflect the browser response accurately.

Thank you!

vejol:
Thanks for your reply! Now I understand what you mean. Unfortunately code in the repository part3-notes-backend is outdated and doesn't match with code examples in course material. The course material if often developed via pull requests, but nobody remembers to add same fixes to actual code repository. It's complicated because the fix should be added to many different brances. However, I will update the code in part3-notes-backend repository soon, in two weeks.

PR #3708, opened 2024-07-10
Update part3a.md

Vopexx:
id to string from the notes array, to make the notes.find(note => note.id === id) in app.get('/api/notes/:id') work properly without string to int conversion

### DIFF
--- a/src/content/3/en/part3a.md
+++ b/src/content/3/en/part3a.md
@@ -187,17 +187,17 @@ const http = require('http')
 // highlight-start
 let notes = [
   {
-    id: 1,
+    id: "1",
     content: "HTML is easy",
     important: true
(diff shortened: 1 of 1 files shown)


alxcd:
Encountered the same problem and have the same solution. Besides, screenshot from material of JSON data is showing ids as strings anyway, so it is logical to define those the same.

image

And it is a better solution than #3720

Vopexx:
Exactly, it solves the problem and is way more consistent with the rest of the part.

> Encountered the same problem and have the same solution. Besides, screenshot from material of JSON data is showing ids as strings anyway, so it is logical to define those the same.
> 
> image
> 
> And it is a better solution than #3720

PR #3711, opened 2024-07-12
Update part1d.md

penguinness:
The buttons somehow do not appear on the screen without "return" (in my case at least).


mluukkai:
the error must be somewhere else, perhaps you had { instead of (

PR #3718, opened 2024-07-15
Update part2e.md

penguinness:
Fix image: Swaziland was renamed to Eswatini in 2018.


vejol:
Pictures are meant to be placed to material repository. Fixed this in another commit.

PR #3720, opened 2024-07-16
Update part3a.md

penguinness:
MAJOR ERROR: The fetched id is a string so it must be converted to a number before being compared with an integer id (in notes).


mluukkai:
resolved by another commit

PR #3739, opened 2024-08-01
Update part0b.md

Divyansh-dotcom:
Changed in DOM : from console tab elements to elements tab


mluukkai:
the word tab is there twice

vejol:
Fixed in other commit.

PR #3745, opened 2024-08-03
eslint config file errors

ksn-ark:
Is currently the standard as in docs https://typescript-eslint.io/getting-started/

Not explicitly importing causes an absolute seizure of crash fit as eslint crumbles in on itself and the text editor becomes unusable (at least in neovim (found out the hard way)).
image

semi is defined at @/semi and not @typescript-eslint/semi

This is some witchcraft caused by the following two reasons: 
The rule for semi has been deprecated - https://typescript-eslint.io/rules/semi/
and moved - https://eslint.style/rules/ts/semi#ts-semi

tested with vs code just to be sure:
image
image


ksn-ark:
Closed because it's a duplicate, sorry

PR #3746, opened 2024-08-03
Update part5e.md

penguinness:
Please make the Cypress chapter legacy. Cypress is not working well with the latest versions of Node (v20.15.0).

Evidence: https://github.com/cypress-io/cypress/issues/25397


vejol:
Thank you for the pull request. We have decided to keep Cypress in the material as an alternative to Playwright, since Cypress is a widely used test framework and is still being actively developed.

PR #3749, opened 2024-08-04
Removed String() frrom addNote function

arkadx13:
Removed String() function from creating id for noteObject. The current result is having an id of type string for newly created notes. This is different from the initial data of notes having data type number for id  on each note.


mluukkai:
id in now a string

arkadx13:
> id in now a string

How come in the github repo it doesnt have that String function? Which one is correct? 
https://github.com/fullstack-hy2020/part2-notes-frontend/blob/part2-2/src/App.jsx

PR #3750, opened 2024-08-04
Update part1a.md - note about StrictMode

domluther:
By default, when using npm create vite@latest part1 -- --template react

main.jsx will look like this

ReactDOM.createRoot(document.getElementById('root')).render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);

Your example on the page doesn't have the <React.StrictMode> tags. I added a note saying this. I asked on Discord and heartsmagic suggested it was probably removed to not confuse learners with strict mode. This note hopefully clarifies that. I am unsure of the exact reason so am happy with any other reason being given.

### DIFF
--- a/src/content/1/en/part1a.md
+++ b/src/content/1/en/part1a.md
@@ -52,6 +52,8 @@ import App from './App'
 ReactDOM.createRoot(document.getElementById('root')).render(<App />)
 ```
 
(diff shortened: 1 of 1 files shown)


vejol:
We do not want to confuse students with unnecessary details. In the first chapter, it is sufficient to replace the contents of the `main.jsx` file with the provided code. Its content is irrelevant to the chapter.

PR #3755, opened 2024-08-05
clearer explanation of state immutability and react rendering (Update…

divywnzh:
… part2d.md)

in reference to the content, the addition now explains (reminds) how modifying the state directly can prevent react from detecting changes, as react relies on object references to detect changes and trigger re-renders.

### DIFF
--- a/src/content/2/en/part2d.md
+++ b/src/content/2/en/part2d.md
@@ -233,7 +233,7 @@ axios.put(url, note).then(response => {
   // ...
 ```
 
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request. The original paragraph provides a good link to the React documentation, which describes well the reasons why mutating state is not recommended in React. It is better to keep the material concise at this stage and provide detailed information through the link. The issue "why state should not be mutated directly" is quite complex and multifaceted at the end.

PR #3756, opened 2024-08-05
Update part3a.md - node --watch instead of nodemon

domluther:
As the goal is to be modern, using node --watch seems a useful sidenote to the nodemon section. You could replace it all if you liked, but I just added that line.


vejol:
Thank you for pointing this out! I fixed this in other commit by removing Nodemon.

PR #3761, opened 2024-08-08
Update part3c.md

nice-Code-90:
Loading environment variables at the beginning of index.js is required.


ntfd0 (approved):
LGTM!
Had an issue going through this section today and this is what was needed! For some reason, the section that follows this (moving to a module) properly mentions adding this require, but this one doesn't. Thank you!

PR #3767, opened 2024-08-10
Update part7c.md

penguinness:
index.html is at the root of the application, not in the public directory.


mluukkai:
no more relevant

PR #3776, opened 2024-08-16
Update part3a.md

sourcyed:
Changed the "important: Boolean(body.important) || false," part to "important: Boolean(body.important)," since the last part was redundant.


vejol:
Thank you for pointing this out. I fixed this in other commit. Idecided to remove Boolean conversion because it seemed to be confusing. I think it is clearer to use `body.important || false` compared to `Boolean(body.important)` although both would work.

PR #3796, opened 2024-09-07
Update part3c.md

RashadAR:
require dotenv to access the mongo url


vejol:
Thank you for the pull request! The Mongo-specific code is isolated in its own module and is only used through the index.js file. Therefore, it is sufficient to load the environment variables only at the beginning of the index.js file. Thus this line is not needed in the note.js file.

PR #3809, opened 2024-09-29
Brian Kernighan is not the developer of the C programming language

aamazigh:
From his Wikipedia page: "Kernighan affirmed that he had no part in the design of the C language ("it's entirely Dennis Ritchie's work")."


bagoviggo (changes requested):
You could also add  "co-author of 'The C Programming Language'" at the end to correct the previous statement without redacting it

PR #3816, opened 2024-10-03
Note for Playwright about unsupported browsers on some operating systems

zeezephyr:
Playwright lacks Webkit support for most Linux distributions except Ubuntu and Debian. Made a generic note here to cover this common case and also cover any other operating systems where browsers may not be supported.


EledenGreen (approved):
yes.
instead of manually calling:
`npm test -- --project chromium` , it would be better to config in the script.

PR #3825, opened 2024-10-09
Update part3c.md

consucv8:
Added line require('dotenv').config(); necessary to use the variables stored in the .env file


vejol:
Thank you for the pull request. About first issue: This issue has now been fixed in another commit with a slightly different solution.

About the second issue: The Mongo-specific code is isolated in its own module and is only used through the index.js file. Therefore, it is sufficient to load the environment variables only at the beginning of the index.js file. This line is not needed in the note.js file.

PR #3826, opened 2024-10-09
Change async thunk call to match createAsyncThunk documentation

zeezephyr:
According to https://redux-toolkit.js.org/api/createAsyncThunk#payloadcreator async thunk functions take a single argument and to pass multiple arguments you'd pass an object instead.


vejol:
Please note that `createAsyncThunk` function is not used in the material. I modified the material a bit and I hope it is clearer now!

PR #3828, opened 2024-10-10
Remove confusing negation in part2d

anttiharju:
Negations increase mental load (one has to flip a value in their head) so it's better to avoid them. There are probably other places in need of similar changes, but this PR updates just the English version of part 2d. I've read through [`part2d.md` with these changes](https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/blob/0ea95fb4a5aa5e0b7733fefbff17907b49c751c1/src/content/2/en/part2d.md) and it appears to remain coherent.

A simple example of the changes:
```diff
axios.put(url, changedNote).then(response => {
-  setNotes(notes.map(n => n.id !== id ? n : response.data))
+  setNotes(notes.map(n => n.id === id ? response.data : n))
})
```

### DIFF
--- a/src/content/2/en/part2d.md
+++ b/src/content/2/en/part2d.md
@@ -204,7 +204,7 @@ const toggleImportanceOf = id => {
   const changedNote = { ...note, important: !note.important }
 
   axios.put(url, changedNote).then(response => {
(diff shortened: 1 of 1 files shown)


anttiharju:
I don't seem to be able to assign a reviewer so pinging

@mluukkai

PR #3832, opened 2024-10-15
updated 3/100.png & typo in part3b.md

EledenGreen:
1. In 100.png,  
   - updated "react dev server" to "vite dev server"  
   - updated from `localhost:3000` to default vite port `localhost:5173`  
  
     image

2. updated sentence structure in "part3b.md"


vejol:
Thank you for the pull request! I fixed this in another commit. The image name had to be changed from the original because not all language versions of the material have been updated to use Vite yet, so the image 100.png still needs to be available in those versions.

PR #3835, opened 2024-10-18
Small bug fix in part8c.md: remove gql before the string literal

patill:
Remove gql before the string literal of the typeDefs constant, since it is used in the client and not the server to to queries.


EledenGreen:
i find having `gql` tag in backend really helps in syntax highlights.  
In part8a.md, using without gql with a string look feels weird.  

> **Without gql:**  

  image

> **With gql:**  

  image

PR #3837, opened 2024-10-19
added 'gql' tag for syntax highlights in GraphQL schema definition

EledenGreen:
added `gql` tag from `graphql-tag` package for syntax highlights in GraphQL schema definition in backend.  

**_Without gql:_**

image

_**With gql:**_

image


vejol:
Thank you for the pull request! I made the changes in a separate commit. Syntax highlighting requires a VS Code plugin to work, so I added brief instructions to the material for the required configuration. I also chose to use a simple type‑indicating comment instead of the `graphql‑tag` library, since this way no extra dependencies need to be installed in the project.

Issue #3843, opened 2024-10-22
Typos in Part 9 (TypeScript), Part C (Typing an Express App) of Full Stack Open


MHammad33:
# Typos in Part 9 (TypeScript), Part C (Typing an Express App)

**Description**:

Hi there! 👋

I came across a couple of typos in **Part 9 (TypeScript)**, specifically in **Part C (Typing an Express App)**. Below are the details:

### 1. Typo in Text Following the Heading "Typing the Request and Response"

- **Location**: Part 9, Part C, immediately following the heading **"Typing the request and response"**
  
  - **Original text**:  
    _"If we hover eg. the parameter res, we notice it has the followng type:"_

  - **Suggested correction**:  
    _"If we hover, for example, over the parameter `res`, we notice it has the following type:"_

### 2. Typo in the Same Section

- **Location**: Part 9, Part C, further down in the section **"Typing the request and response"**

  - **Original text**:  
    _"We could and propably should give a proper type as the type variable."_

  - **Suggested correction**:  
    _"We could and probably should give a proper type as the type variable."_

### Additional Notes:

- Both typos are minor but could lead to confusion for readers.
- Happy to submit a pull request if that would be helpful!

Thanks for maintaining such a fantastic course. Let me know if you need any further clarification!

Vikram-Singh0:
opened pull request pls merge

mluukkai:
fixed

PR #3846, opened 2024-10-27
Fixed two typo errors in part-9-c #3843

Vikram-Singh0:
So i fixed two typos in the part 9-c one was simple spelling mistake and second leads to confusion for readers which i corrected.

### DIFF
--- a/src/content/9/en/part9c.md
+++ b/src/content/9/en/part9c.md
@@ -817,7 +817,7 @@ The response is what we expect it to be:
 
 So far we have not discussed anything about the types of the route handler parameters. 
 
-If we hover eg. the parameter _res_, we notice it has the followng type:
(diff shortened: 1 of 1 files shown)


src/content/9/en/part9c.md
```diff
@@ -817,7 +817,7 @@ The response is what we expect it to be:
 
 So far we have not discussed anything about the types of the route handler parameters. 
 
-If we hover eg. the parameter _res_, we notice it has the followng type:
+If we hover, for example, over the parameter res, we notice it has the following type:
```
EledenGreen, line 820:
```suggestion
If we hover, for example, over the parameter _res_, we notice it has the following type:
```

EledenGreen (changes requested):

mluukkai:
fixed elsewhere

PR #3855, opened 2024-11-04
Updated response type of router.get('/') in Part9c.md

EledenGreen:
1. **Updated Response type**.
  `diaryService.getNonSensitiveEntries()` return type is : `NonSensitiveDiaryEntry[]`. 
  [reference](https://fullstackopen.com/en/part9/typing_an_express_app#utility-types): 
  `const getNonSensitiveEntries = (): NonSensitiveDiaryEntry[] => {
  return diaries;
};`
So Response type of the route: `router.get('/',`  should be `NonSensitiveDiaryEntry[]`

2. updated sentences.

### DIFF
--- a/src/content/9/en/part9c.md
+++ b/src/content/9/en/part9c.md
@@ -842,7 +842,7 @@ import { Response } from 'express'
 
 // ...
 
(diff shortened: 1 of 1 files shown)


mluukkai:
fixed in new material

PR #3856, opened 2024-11-04
update Ex: 9.27 instruction to use Zod library

EledenGreen:
**Updated instruction in Exercise: [9.27](https://fullstackopen.com/en/part9/grande_finale_patientor#exercises-9-23-9-30) .**

instead of writing custom validator : 
```
const parseDiagnosisCodes = (object: unknown): Array<Diagnosis['code']> =>  {
  if (!object || typeof object !== 'object' || !('diagnosisCodes' in object)) {
    // we will just trust the data to be in correct form
    return [] as Array<Diagnosis['code']>;
  }

  return object.diagnosisCodes as Array<Diagnosis['code']>;
};
```
using Zod library instead:
```
const DiagnosisCodesSchema = z
  .array(z.string())
  .optional()
  .transform((codes) => (codes ? codes : []));
  ```

### DIFF
--- a/src/content/9/en/part9e.md
+++ b/src/content/9/en/part9e.md
@@ -402,17 +402,18 @@ Remember that we have different kinds of entries in our app, so our backend shou
 
 In this exercise, you quite likely need to remember [this trick](/en/part9/grande_finale_patientor#omit-with-unions).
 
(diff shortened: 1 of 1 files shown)


mluukkai:
fixed in new material

PR #3869, opened 2024-11-17
Update part6a.md

shu1513:
Starting from ESLint v8.21.0, ESLint introduced a new configuration format (eslint.config.js) as an alternative to traditional configuration files like .eslintrc.cjs.


vejol:
Thank you for your pull request. The material has now been updated in a separate commit. There is no need to do changes to `eslint.config.js` file anymore at this point of material.

PR #3871, opened 2024-11-17
Update part6a.md

shu1513:
come on...don't just let the course go the further the chapters.


mluukkai:
Eg this is purposefully left out. The student is supposed to notice this kind of trivial things by him/herself.

PR #3873, opened 2024-11-18
Update part6a.md

shu1513:
Remember that I said the quality of the course has been steadily declining after part 2? Here is a typical example. As someone who is learning it for the first time, it is reasonable, if not a common-sense to make clear to the new learner WHERE the code goes. App.jsx? main.jsx? or a seperate file then later to be imported? An increasing number of mistakes like this in the later chapters suggest that as time went on, the author just let the course go.

### DIFF
--- a/src/content/6/en/part6a.md
+++ b/src/content/6/en/part6a.md
@@ -748,7 +748,7 @@ Your application can have a modest appearance, nothing else is needed but button
 
 ### Uncontrolled form
 
-Let's add the functionality for adding new notes and changing their importance:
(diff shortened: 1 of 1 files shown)


shu1513:
I understand. I have also heard that this is closer to 'real-world' scenario.

Overall the course has great concepts. In fact, the course covers excellent and important materials. It is my experience some parts can use more clarifying details which will make it less unnecessarily confusing.

Thank you Matt,
This course overall is very valuable.
________________________________

PR #3875, opened 2024-11-19
Update part6a.md

shu1513:
The edits should be self-explanatory.


mluukkai:
will not accept his. at this point it should be pretty clear for a student that the imports are needed

PR #3888, opened 2024-11-28
Update part6c.md

shu1513:
To make it clear which directory to put in the new file and the service directory. Without this clear instruction, I put the service directory in the root file the first time I was following through the course material.


mluukkai:
this should be clear from the context since **all** the source code in the course goes under src, will not merge

PR #3898, opened 2024-12-05
Update part6d.md

shu1513:
Using object instead of an array, this way you won't need any of the helper functions or any of the adjustments in other files. Furthermore, you avoid digging yourself into a hole(a possible future mistake) when you use object keys instead of array indexing. I hope you can see what I am trying to do here. I was indeed surprised when I saw when you used array.  Also, the helper functions, even though I know what they do, in the future when someone else try to understand this code, it would require whomever reads this code to go around an additional circle to understand the code, instead of the directness of not using helper functions even if you use old codes. So maybe helper functions are not always as helpful, sometimes they can be counter-helpful and make the code more confusing harder to undersatnd.

### DIFF
--- a/src/content/6/en/part6d.md
+++ b/src/content/6/en/part6d.md
@@ -570,7 +570,7 @@ const App = () => {
   const [counter, counterDispatch] = useReducer(counterReducer, 0)
 
   return (
-    <CounterContext.Provider value={[counter, counterDispatch]}>  // highlight-line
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request! I agree with you, the object syntax is a bit clearer in this situation. I fixed this in a separate commit as it was needed to re-write part of the material.

PR #3899, opened 2024-12-05
Update part6d.md

shu1513:
This link no longer works.


vejol:
Thank you for the pull request! The link seems to work now. I have noticed that TanStack documentation pages are a bit unreliable speaking of availability, as every now and then they may be unreachable.

PR #3913, opened 2024-12-11
Update part2d.md

standingpaul:
This might not be the best solution but, I got stuck not understand why my error was not being caught. It was because I already put the catches into the Notes.jsx service script, consequently the catch in App.jsx did not catch anything and I had an array of notes with an undefined item.


vejol:
Thank you for pointing this out! I fixed this in other commit as the material needed a bit more refinement to be clear.

PR #3916, opened 2024-12-12
Update part7c.md

shu1513:
I could be wrong. I think the index.html file is in the root directory? The public directory only has vite.svg.


mluukkai:
no more relevant

PR #3919, opened 2024-12-14
Update part7d.md

shu1513:
Why do I make pull requests like this? On seemingly small details that students "should" already know, or can easily find out themselves? The reason is that as we are learning, when we have to deviate away from the main flow of logic, to tasks such as finding out which file it belongs to, importing something or not, or "small" details where to find something, this extra step, deviates us from the flow of the logic of the course. I have heard a study that once a person gets distracted like this, it takes 20 minutes to get back to the focus of the main course. So for a more effective and efficient learning, it is best to take care of these details. Not to distract the students away from the main flow of the logic. Rather than just saying "they should know this by now". The key is in the details.


mluukkai:
this is so obvious that will not merge

PR #3920, opened 2024-12-14
Update part7d.md

shu1513:
What "open file" functionality of my browser? What browser? Safari? Chrome? Do they offer that functionality? how to use it? I don't remember seeing any previous explanation of  "open file" functionality of any browser in the course so far. It is like casually saying you can simply use Postgres to handle database - without explaining what it is or how to use it.


mluukkai:
google will tell. and once more, it is intentional to make students think and solve small problems by themselves

PR #3925, opened 2024-12-17
Update part7e.md

shu1513:
Can we do a chapter on serverless Matti?


mluukkai:
pr is not right media for this kind of propositions

PR #3926, opened 2024-12-17
Update part1d.md, part "A function that returns a function"

kaelwi:
At the end should be setValue() instead of setToValue in the event handler.


vejol:
There is no mistake in original code.

PR #3927, opened 2024-12-17
Update part1d.md, part "Passing Event Handlers to Child Components"

kaelwi:
Use setValue instead of setToValue in event handlers.


vejol:
There is no mistake in original code. setToValue function is defined in App.jsx in the previous chapter.

PR #3928, opened 2024-12-19
Update part3c.md

standingpaul:
I think it should be mentioned somewhere nearby where I put the comment that you need 'require("dotenv").config();' to be able to access the env variable to test the database working with the backend code. It is mentioned only a bit further on in the text, but since it says you can verify in the browser, which is not possible without this reference.


vejol:

PR #3933, opened 2024-12-26
Update part6a.md

fixprogram:
We won't clear the input with event.target.note.value = '' because it's just an event object. 
If we're talking about uncontrolled components then I suppose we want to use refs here.

P.S We should also do something with description below about getting the input value by it's name


vejol:
Thank you for your pull request. Actually, the input can be cleared with `event.target.note.value = ''`. I'm closing this now.

PR #3940, opened 2025-01-04
Update part1d.md

amr112004:
making more calrifications about useState


vejol:
This is not relevant at this point in the material because useState has been introduced in detail in an earlier chapter.

PR #3941, opened 2025-01-05
Fix destructuring of props for Button in Part 1

loveleenc:


vejol:
Fixed in other commit.

PR #3945, opened 2025-01-07
Update part3d.md

sudo-rickroll:
Added a Note at the "vscode eslint plugin" part to ensure that it works. Without the changes to settings.json, it does not work.


vejol:
Thank you for the pull request! This is a bit too specific to be added to the course material. The eslint.validate setting should not exist automatically in the settings.json file, nor is it added during the installation of the VSCode ESLint plugin. Therefore, this instruction applies only in very exceptional cases. It is also beneficial for learning that people have to figure out their own environment configuration issues themselves, rather than trying to provide a ready-made solution for everything.

PR #3948, opened 2025-01-09
Fix docker container ls -a command

sebazai:


sebazai:
Closing, as same is later in material and seems to be conscious decision :)

PR #3959, opened 2025-01-16
Update part9c.md

shu1513:
Both I and my IDE were wondering where did "z" came from.


mluukkai:
fixed in new material

PR #3965, opened 2025-01-20
part2e 'kebab case' => 'kebab-case'

aleichtm:


vejol:
Thank you for the pull request. In this case, "kebab case" is the name of the naming convention being used, so it is written without a hyphen. Its meaning is clarified by the examples in the text, such as font-size vs. fontSize. Therefore, I do not accept this change.

PR #3969, opened 2025-01-24
Update info in part2a.md, exercise 2.1

juusosar:
The example solution for courseinfo for exercise 2.1 is currently found from the file 'App.js' in the submission system. Also the reference to 'kurssitiedot' might be deprecated since the submissions site has solution folders in English?


vejol:
Thank you for pointing this out! Could you make a little adjustment, please? File name is actually App.jsx

PR #3970, opened 2025-01-24
Specify major ESLint version

watteja:
ESLint 9 only supports flat config files. Package `eslint-plugin-react-native` does not. Therefore, it's necessary to install ESLint v8.


watteja:
I've just finished part 10, and I believe this is just a first instance of a wider issue regarding package compatibility.

I saw in later materials that usage of `--legacy-peer-deps` flag is recommended. For me personally, I also used `npx expo install` often instead of `npm install` directly, which ensures that the installed package will be compatible with the used Expo SDK version.

As I don't know enough about package management in Node (or Expo), I am closing this PR and letting more experienced folk handle it at the appropriate scale =)

PR #3971, opened 2025-01-24
Update part5b.md

sudo-rickroll:
Proptype added for 'children' inside Togglable.jsx to prevent eslint errors on it.


vejol:
Thank you for the pull request! Unfortunately, PropTypes is deprecated and will be removed in React version 19. The section related to it will be removed from this material during the update, so I will not accept this pull request.

PR #3974, opened 2025-01-29
Update part2b.md

SilverZer0:
The link is broken. I replaced it with a link to an archived version on web.archive.org


vejol:
Fixed in other commit.

PR #3975, opened 2025-01-29
Dead link to JoshBritz object equality, update to MDN docs

ConnorWBrown:
The link out to JoshBritz.co about object equality in JavaScript is dead, looks like the site is down.

The following article from Mozilla seems likely to cover the same information:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness


rlarg:
I'd also recommend this link: https://adripofjavascript.com/blog/drips/object-equality-in-javascript.html as it offers a few approaches to finding a solution.

PR #3977, opened 2025-01-31
Update part13a.md

shu1513:
ssl is no longer supported. I actually don't know what is the correct way but deleting "ssl: {}" does make it work.

### DIFF
--- a/src/content/13/en/part13a.md
+++ b/src/content/13/en/part13a.md
@@ -301,12 +301,7 @@ Note: if you use Heroku, you might need an extra option in connecting the databa
 ```js
 const sequelize = new Sequelize(process.env.DATABASE_URL, {
   // highlight-start
-  dialectOptions: {
-    ssl: {
-      require: true,
-      rejectUnauthorized: false
-    }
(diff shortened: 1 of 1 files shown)


watteja:
A bit earlier in the module, it's mentioned that `dialectOptions` might be necessary on Heroku specifically.

So I suspect this is an artifact from that environment. I can confirm however that this won't work on Postgres server running in Docker container for example, since SSL is not supported out of the box. [There is a way to rebuild the image using certificate-based authentication and OpenSSL](https://dev.to/danvixent/how-to-setup-postgresql-with-ssl-inside-a-docker-container-5f3), but I really think that's outside of the scope of this module, even for the standards of this course.

So I agree that the line should be left to simple `const sequelize = new Sequelize(process.env.DATABASE_URL)`, especially now that Heroku doesn't offer a free tier and Fly.io doesn't support managed Postgres.

Just bear in mind that you should changed it consistently in all occurrences or none at all.

mluukkai:
fixed in another commit

PR #3991, opened 2025-02-15
Update part3c.md

Rehnyk:
Example cannot read environmental variables at this point. It's missing 'require('dotenv').config()' and application breaks. It is explained in next section and it becomes redundant.


vejol:

PR #3995, opened 2025-02-18
typo fix in part 12c.md

EledenGreen:


vejol:
Thank you for the pull request! The sentences are actually intentionally placed together. "Avoid bugs that appear only in the production environment" is the reason why it is wise to keep the environment similar between development and production. I am closing this pull request now, but please do not hesitate to reopen it if you would like to address this further.

PR #3996, opened 2025-02-19
Update osa5b.md

astranero:
Deprekaatio tulossa forwardRef:ille. 

https://react.dev/reference/react/forwardRef
Syyt: https://youtu.be/m4QbeS9BTNU


vejol:

PR #4000, opened 2025-02-23
Update part3b.md

JavIregui:
Adding [Railway](https://railway.com) to the list of services for deploy. Fly.io is no longer free but Railway is a really easy to use alternative that allows you to use up to 5$ for free each month. More than enough for basic projects.

### DIFF
--- a/src/content/3/en/part3b.md
+++ b/src/content/3/en/part3b.md
@@ -103,6 +103,7 @@ There are also some other free hosting options that work well for this course, a
 
 Some course participants have also used the following services:
 
+- [Railway](https://railway.com)
 - [Replit](https://replit.com)
 - [CodeSandBox](https://codesandbox.io)
 


vejol:
Thank you for the pull request! Do you have personal experience with Railway? At first glance, it seems that a monthly fee of $5 is charged regardless, but you get a $5 discount on usage-based costs each month:

image

JavIregui:
I've used it many times for college projects and it's really simple to use. It has templates where you can choose your stack and it will configure everything for you. You just have to connect it to a git repository.

Regarding the cost, as you can see in the images below, you can set a limit for the Hobby plan ($10 in my case but you can set it to $5 really easily) and it calculates how much you have used already and the expected cost at the end of the month. Then they give everyone a $5 discount so, as long as you haven't used more than that, it's always free.

It seems complicated but you don't have to configure any of it, the discount works completely automatically. You can configure the limit just in case if you really want to but it only takes a second.

Captura de pantalla 2025-02-23 a las 21 03 00
Captura de pantalla 2025-02-23 a las 21 02 14

vejol:
Thank you for the detailed response! It sounds like you have a different plan compared to what they currently offer. It seems that your Hobby plan costs $0/mo + usage - $5 gift. However, Railway's website states that the current price is $5/mo + usage - $5 gift. Do you understand what I mean? Maybe I need to try the service personally to be sure about it?

JavIregui:
I understand. I did it a long time ago and maybe the plans have changed and I still have the old one. I think I remember it being confusing at first for me too and I also thought that it worked like that but it turned out to be free every month because it applies the discount to the final bill. Maybe it would be better to test it with a new account to be sure.

I've also checked my subscription and I have the regular Hobby one and in the details it also says "Includes $5 of usage monthly" but as you can see in the previous images it always ends up being free.

I hope this helps!

Captura de pantalla 2025-02-25 a las 19 07 07

vejol:
Okay, I will look into it. Thank you very much for your help!

PR #4002, opened 2025-02-25
Correct information about env context availability

watteja:
To the best of my knowledge, the removed information was not correct. I have tried myself to access environment variable with `github.env`, and it doesn't work. I then went online only to find advice such as [this one](https://stackoverflow.com/questions/73558652/github-actions-how-to-use-environment-variable-at-job-level), which worked for me and which I changed in the materials.

[The officiall documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#context-availability) also doesn't mention availability at the `jobs` level, though I may be reading that incorrectly.

### DIFF
--- a/src/content/11/en/part11d.md
+++ b/src/content/11/en/part11d.md
@@ -292,6 +292,8 @@ env:
 jobs:
   job1:
     # rest of the job
+    outputs:
+      condition: ${{ env.CONDITION }}
     steps:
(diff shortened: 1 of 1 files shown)


mluukkai:
not relevant after some changes

PR #4003, opened 2025-02-25
Suggest different Discord notification action

watteja:
While the exercise description advises picking *any* action from the Marketplace that does the job, I found the inclusion of previous action to be misleading, as it is currently broken even for most basic functionality.

After using embarrassing amount of time to make sure the problem was not with me, I found an alternative that can create notifications identical to the one included in the screenshots (also much more robust, given its simplicity of implementation). See my own notifications for proof

custom_notifications

*Ignore identical commit times for both notifications, that is not action's fault, but mine. I'm aware I shouldn't --force push, but that's outside of the scope of this PR*

### DIFF
--- a/src/content/11/en/part11e.md
+++ b/src/content/11/en/part11e.md
@@ -39,7 +39,7 @@ Register now to Discord if you have not already done that. You will also need a 
 
 #### 11.18 Build success/failure notification action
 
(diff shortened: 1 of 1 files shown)


mluukkai:
fixed in new material

PR #4004, opened 2025-02-25
change the way to get message from onError option the useMutation hook

barbarred:
The way the apolloError error message is accessed has been modified


vejol:
Thank you for your pull request! This was fixed in another commit.

PR #4006, opened 2025-02-27
Update: part3c.md with require('dotenv').config()

waikoo:
To avoid confusion in case one would get the error that the url variable from mongoose.connect(url) is undefined, like I did.


vejol:

PR #4007, opened 2025-03-01
change dockerignore by gitignore

WilmarDeML:
I also propose to update the example application because they are using the script with heroku, I think it should be with Fly.io or Render.


vejol:
Thank you for the pull request! The material intentionally mentions the `.dockerignore` file. It specifies which files Fly.io includes in the Docker image used for deployment. Fly.io does not deploy files from the GitHub repository but rather from the local repository. Therefore, this change is unnecessary.

The example application code is also expected to be updated soon!

PR #4009, opened 2025-03-05
Update part12b.md

shu1513:
The "/statistics" path is within the "/todos" path if I am not mistaken? because I made the GET "/statistics" path it within the todos.js


mluukkai:
this is not right

PR #4010, opened 2025-03-05
Update part4b.md

kemes:
Define assert in the supertest -chapter was missing. This change adds the definition to the code example.


vejol:
Thank you for pointing this out! This was fixed in another commit.

PR #4013, opened 2025-03-07
Fix useServer import path for graphql-ws 6.0.4

JoeHsiao:
npm installs "graphql-ws": "^6.0.4" in package.json. The new path to import useServer is 'graphql-ws/use/ws' instead of 'graphql-ws/lib/use/ws'.


vejol:

PR #4017, opened 2025-03-15
Update part1c.md

Adhin-sujo:
ReactDOM.createRoot is discontinued. Use createRoot only to fix the issue. example:
import { createRoot } from 'react-dom/client'
createRoot(document.getElementById('root')).render(
  <StrictMode>
    <App counter={counter}/>
  </StrictMode>,
)

### DIFF
--- a/src/content/1/en/part1c.md
+++ b/src/content/1/en/part1c.md
@@ -200,7 +200,7 @@ import App from './App'
 
 let counter = 1
 
-ReactDOM.createRoot(document.getElementById('root')).render(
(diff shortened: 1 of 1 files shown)


vejol:
Hi! Thank you for the pull request. What do you mean when you say that ReactDOM.createRoot is discontinued? 

This is just an alternative way to do the import:
```jsx
import ReactDOM from 'react-dom/client'
ReactDOM.createRoot(...)
```
This method has been used consistently throughout the material, so changing it would be laborious.

PR #4019, opened 2025-03-19
Suggest further optimization example for Button component refactoring

meeroslaph:


mluukkai (changes requested):
; should be removed

meeroslaph:
> ; should be removed

[Removed](https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/pull/4019/commits/7f41dd04aadfdc6ead8e8126be816a5aaffd0f3a).

PR #4021, opened 2025-03-20
Update 1.7 exercise description

meeroslaph:
The description mentions `bad: -1`. Is it possible to get a negative number of feedbacks?


vejol:
Thank you for the pull request! This is actually meant to provide instructions for calculating the average of the feedback. The average ranges between -1 and 1, meaning the value of negative feedback is -1. I added a note to the task description clarifying that it specifically refers to the values of the feedback.

PR #4022, opened 2025-03-24
Update part8a.md

WilmarDeML:
gql is undefined


vejol:
Fixed in another commit (PR #4012)

PR #4027, opened 2025-03-27
Update part11c.md

shu1513:
The 'fly auth token' command is deprecated. Use 'fly tokens create' instead.

### DIFF
--- a/src/content/11/en/part11c.md
+++ b/src/content/11/en/part11c.md
@@ -84,7 +84,7 @@ Before going to the automated deployment, we shall ensure in this exercise that 
 So, create a new app in Fly.io. After that generate a Fly.io API token with the command
 
 ```bash
-flyctl auth token
+fly tokens create
 ```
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request! Did you test the code you proposed? It seems that in the new command, the token type needs to be specified separately, so the correct command might be something like: `fly tokens create deploy`.

The command `fly auth token` still seems to work, even though it is deprecated. The new command generates a very different-looking token, and it should be tested to see if it works the same way as the previous token or if additional changes are needed in the course material.

shu1513:
You are right. I actually didn't test it. I used the deprecated token. the fly.io website uses this command "fly tokens create deploy -x 999999h". when I used the fly.io website command following their tutorial, it worked. here is the page:https://fly.io/docs/launch/continuous-deployment-with-github-actions/

vejol:

PR #4030, opened 2025-04-10
Update part11d.md

leburgeon:
Broken link fix. Hovering previously displayed a question mark.


vejol:
Thank you for the pull request. This is not actually a link but an `abbr` element that explains the meaning of the abbreviation. Hovering over it displays a question mark and the meaning of the abbreviation. Its use improves the accessibility of the page.

I also don't think it's necessary to add a separate link to GitHub's documentation here, as the material already explains the same content about Pull Requests as GitHub's documentation.

leburgeon:
Thank you for the explanation!

PR #4034, opened 2025-04-13
Update part5b.md forwardRef

Badiealili:
Updated the code in the section "References to components with ref" section and added explanation relative to the deprecated function forwardRef (deprecated as of React 19) and the new way to access refs in react components as props.


vejol:

PR #4045, opened 2025-04-21
Update part4b.md

WingRa7:
Add 'notes' just to make it clear, and not implied, which app we're working on.


mluukkai:
that should be obvious from the context

PR #4052, opened 2025-04-28
Update part3c.md : typo

RuilinYang-beta:
Corrected a small typo: 
"If you define a model with the name Person, mongoose will automatically name the associated collection as people."  The collection name should be "persons".


vejol:
It's not actually a typo; the name of the collection will be "people." You can try it yourself by following the example in the material.

PR #4055, opened 2025-05-11
Update part8e.md

eedrxs:
Updated `graphql-ws` import to match current version. Following the tutorial as it is currently, the code breaks because it cannot find the export path:

`const { useServer } = require("graphql-ws/lib/use/ws")`

image

I checked the `graphql-ws` docs and found the import path was different. Updated my code and everything works again:

`const { useServer } = require("graphql-ws/use/ws")`


vejol:

PR #4056, opened 2025-05-11
Update part8e.md

eedrxs:
Update use of `pubsub.asyncIterator` to `pubsub.asyncIterableIterator`.

Following the tutorial as it is currently, you get this error when you try to listen for subscriptions:

image

The current code makes use of `pubsub.asyncIterator`, but this method no longer exists and has been changed to `pubsub.asyncIterableIterator` in the latest `graphql-subscriptions` version. Updating the code with this makes everything work fine:

image


vejol:

PR #4061, opened 2025-05-18
Update part2c.md

miguelc83:
docs(part2c): add optional improvement using concurrently to run React and json-server simultaneously

This commit adds an optional section explaining how to use the `concurrently` tool to run both the React app (with Vite) and the `json-server` mock API with a single command (`npm run dev:full`).

While running the processes separately is pedagogically valuable, using `concurrently` streamlines the development workflow and aligns with practices commonly used in professional environments.

### DIFF
--- a/src/content/2/es/part2c.md
+++ b/src/content/2/es/part2c.md
@@ -226,12 +226,6 @@ y haciendo una pequeña adición a la parte <i>scripts</i> del archivo <i>packag
 }
 ```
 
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request! I appreciate the effort you put into this. However, I believe this does not serve the purpose of the course. Many students have difficulty understanding what happens where, so starting the processes in separate terminal windows is appropriate at this stage of the course. For this reason, I will not accept this pull request now, so that the course material remains concise and focused on the essentials.

PR #4066, opened 2025-06-06
update part6a.md: instructions for students to use `jest` functions globally

jchanke:
add: instructions to configure `jest` globals for vite: ^6.3.5

the existing notes pre-suppose that a file `.eslintrc.cjs` exists, which doesn't come with newer versions of vite. 

these changes guide students using a newer version of vite to configure their project so jest globals (describe, test, etc.) can be used without importing.

thanks! :)


vejol:
Thank you for your pull request. The material has now been updated to use Vitest so this is not relevant anymore.

PR #4076, opened 2025-06-25
Update part9c.md: update zod to v4

jchanke:
updated zod syntax for the latest version of zod (v4).

* updated `z.nativeEnum() --> z.enum()` ([link to docs](https://zod.dev/api?id=enums))
* updated `z.string().date() --> z.iso.date()` ([link to docs](https://zod.dev/api?id=iso-dates)


mluukkai:
fixed in new material

PR #4081, opened 2025-07-01
Update part3a.md notes backend post method empty content handling

LCUnger:
I've been following along the course in typescript, note that I'm really not experienced with it. But I did a course on it just before this one and wanted to get some experience with it.

This part of the code wouldn't work (line 672 gave an ts error, see error below). All ai tools I tried also didn't understand it. I think it's because the Type 'Response' which you would give the response variable if you're working in typescript doesn't allow for returning it or the post function type doesn't like it. By forcing any on the req and res I got it working but this felt of because everyone tells me you shouldn't use any in ts. From just intuition it felt weird returning a response method. So I tried separating the return to after editing the response status. This fixed everything.

I know this part of the course is not made for typescript, but I figured if typescript likes it, it's may be a better way to write the code.

Error: src/index.ts:62:24 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '(req: Request, res: Response) => express.Response<any, Record<string, any>> | undefined' is not assignable to parameter of type 'Application<Record<string, any>>'.
      Type '(req: Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>, res: Response<any, Record<string, any>>) => Response<...> | undefined' is missing the following properties from type 'Application<Record<string, any>>': init, defaultConfiguration, engine, set, and 63 more.

62 app.post('/api/notes', (req: Request, res: Response) => {
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@types/express-serve-static-core/index.d.ts:168:5
    168     (path: PathParams, subApplication: Application): T;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    The last overload is declared here.

### DIFF
--- a/src/content/3/en/part3a.md
+++ b/src/content/3/en/part3a.md
@@ -670,9 +670,10 @@ app.post('/api/notes', (request, response) => {
   const body = request.body
 
   if (!body.content) {
-    return response.status(400).json({ 
+    response.status(400).json({ 
       error: 'content missing' 
     })
+    return;
   }
 
   const note = {


vejol:
Thank you for the pull request. This kind of syntax is used several times in different parts of the course material, and I think it is a clear and functional choice for these situations. Therefore, I will not accept this pull request.

PR #4084, opened 2025-07-05
Update part12c.md

rksab:
Using both nginx:1.25-alpine and nginx:1.20.1. Better to update to  nginx:1.25-alpine

### DIFF
--- a/src/content/12/en/part12c.md
+++ b/src/content/12/en/part12c.md
@@ -486,7 +486,7 @@ services:
   app:
     # ...
   nginx:
-    image: nginx:1.20.1
+    image: nginx:1.25-alpine
     volumes:
       - ./nginx.dev.conf:/etc/nginx/nginx.conf:ro
     ports:


ziirou (changes requested):
Maybe update the version/image to the container list too.

```bash
$ docker container ls
CONTAINER ID   IMAGE              COMMAND  PORTS                   NAMES
a02ae58f3e8d   nginx:1.25-alpine  ...      0.0.0.0:8080->80/tcp    reverse-proxy
5ee0284566b4   hello-front-dev    ...      0.0.0.0:5173->5173/tcp  hello-front-dev
```

EDIT: I'm opened my PR again as there were more `nginx:1.20.1` used and found out that `bash` is not included in Alpine-based images.

rksab:
I'm not sure why 2 diff versions are being used. As a novice, it's a bit confusing. If 1.20.1 is being used during development for better compatibility, some text needs to be added.

mluukkai:
fixed elsewhere

PR #4088, opened 2025-07-12
docs: add warning about express-async-errors peer-dependency on Expre…

vaibh123540:
Added a conspicuous markdown warning below the `npm install express-async-errors` instruction  to inform Express 5 users that the package currently declares a peer dependency on Express 4.


vejol:
Thank you for the pull request! The material has now been updated to use Express 5. This was fixed in a separate commit, as the update required changes in several places throughout the material.

PR #4089, opened 2025-07-15
Update part5b.md

PabloDiazDev:
React 19 and ref as a prop instead of forwardRef.


vejol:

PR #4092, opened 2025-07-17
To maintain consistency, the description of the fourth component has been removed.

asp8964:
* Deleted the specific description of the fourth component mentioned in the example.
* Ensured the text now consistently refers to and describes three components, matching the initial statement.


vejol:
Thank you for the pull request. The text intentionally lists four different candidates for components to be refactored. It's good that students have to think a bit themselves about what they think should be refactored. Therefore, I won't accept this pull request.

PR #4093, opened 2025-07-24
Update part11: Change ubuntu github-runner version to latest

wesenbergg:
Github runner does not support ubuntu-20.04 image anymore. The alternative is to use ubuntu-22.04 or ubuntu-24.04 (ubuntu-latest).

Screenshot 2025-07-24 at 20 10 52
[Screenshot of available runners](https://github.com/actions/runner-images)


mluukkai:
fixed in another commit

PR #4096, opened 2025-07-29
Part 3D: change eslint configuration

AlainLobato:
ESlint change it configuration so i had a little problems when i was configuring the eslint file.

Just update links and configuration in the config file to configure correctly.


vejol:
Thank you for the pull request! The ESLint section in the Spanish material does indeed appear to be outdated. The updated version can be found in the English material: https://fullstackopen.com/en/part3/validation_and_es_lint#lint 

I noticed that the changes you suggested are not fully consistent with the updated material. Would it be possible for you to update the Spanish material to match this English material, please?

vejol:
Since I have not received a response, I will merge these changes now. However, the English material contains a more up-to-date version of these ESLint configurations, so it would be advisable to update the Spanish version accordingly at some point.

AlainLobato:
I'm sorry I was busy, but I will do that as soon as i can. Thanks and have a nice day.

- DevLine

PR #4100, opened 2025-08-05
Update part4b.md

tokisuno:
Removed mention of the ``express-async-errors`` library, as it is no longer required.


vejol:
Thank you for the pull request! The material has now been updated to use Express 5. This was fixed in a separate commit, as the update required changes in several places throughout the material.

tokisuno:
Ah okay, that's great to hear. Thank you!

PR #4105, opened 2025-08-12
update zh/part5c.md

Juemuren:
I think I have fixed all the errors in zh/part5c.md . There is a lot of commits. Do I need to create a new PR to make all the changes into one commit?


vejol:
Thank you for the pull request! Of course, it looks a bit neater if small changes like this are in a single commit rather than several. However, it’s not a problem that there are multiple commits here, so I’ll approve this now.

PR #4120, opened 2025-08-23
part8e: update import path for useServer

WingRa7:
Previous import path for `useServer` `graphql-ws/lib/use/ws` leads to runtime error `useServer is not a function`

`graphql-ws` v6 updated path to `graphql-ws/use/ws`


vejol:

PR #4122, opened 2025-08-28
Update part9c.md - removed unused @types/eslint__js

Uimalakki:
ESLint 9 moved to ESM and ships its own types. The @types/eslint__js installed here is basically a ghost (and with it TS complains it can’t find eslint__js).


mluukkai:
removed from new material

PR #4126, opened 2025-08-31
Update part9b.md: @types/eslint__js deprecated

WingRa7:
@eslint/js provides its own type definitions, so you don't need @types/eslint__js installed!


mluukkai:
done in new material

PR #4134, opened 2025-09-03
Update part12c.md

ziirou:
Added a fix for wget connecting problem.

I was stuck with this for quite long and it seemed that others have had similar problems (in Discord channel).

### DIFF
--- a/src/content/12/en/part12c.md
+++ b/src/content/12/en/part12c.md
@@ -354,6 +354,13 @@ writing to stdout
       ...
 ```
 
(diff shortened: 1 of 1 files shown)


ziirou:
Found this too from the Discord channel: https://github.com/vitejs/vite/commit/bd896fb5f312fc0ff1730166d1d142fc0d34ba6d
So the `allowedHosts` has been added to server options for Vite v7.1.4.

WingRa7:
Came across the same problem too.

I'm thinking if we could use `host:true` to avoid having to then use the --host flag, and the pros/cons of using `allowedHosts: true` vs `['app']`.

```js
  server: {
    host: true,
    allowedHosts: true,
  },
  ``

mluukkai:
fixed elsewhere

PR #4136, opened 2025-09-04
Update part12c.md

ziirou:
Use the same Nginx version (`nginx:1.25-alpine`) as earlier (Part 12c: Using multiple stages) instead of `nginx:1.20.1`.

`bash` is not included in Alpine-based images (like `nginx:1.25-alpine`).
Got this error:
> OCI runtime exec failed: exec failed: unable to start container process: exec: "bash": executable file not found in $PATH: unknown

Use `sh` instead of `bash`: `docker exec -it reverse-proxy sh`

### DIFF
--- a/src/content/12/en/part12c.md
+++ b/src/content/12/en/part12c.md
@@ -525,7 +525,7 @@ To help us, Docker Compose has set up a network when we ran _docker compose up_.
 Since we are inside the container, we can also test the DNS! Let's curl the service name (app) in port 5173
 
 ```html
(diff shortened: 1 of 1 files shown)


ziirou:
There seems to exist a similar PR already: #4084
So I'll close this one.

ziirou:
I'm opening this again as there were more `nginx:1.20.1` used and found out that `bash` is not included in Alpine-based images.

mluukkai:
fixed elsewhere

PR #4143, opened 2025-09-07
Update part9c.md: zod deprecated syntax

WingRa7:
new zod syntax for date and enum


mluukkai:
fixed in new material

Issue #3263, opened 2023-11-09
findByIdAndUpdate doesnt validate required paths even if runValidators is true, not suitable for PUT


gregorianrants:
## The Issue

I have raised this issue before on discord, and i think some things were modified but i still think there are problems with using

findByIdAndUpdate for a put request.

even when you use {runValidators: true}

it still only runs validation on the paths we supply to it.
a put request should replace a document with the data provided.
however if we dont provide a value for a **required** path in the update, it wont run the required validator on the path that has no value.  Only the paths provided will be replaced and the path that wasnt supplied will still have its previous value.

as pointed out in the mongoose docs

https://mongoosejs.com/docs/validation.html#update-validators

it also says in the mongoose docs

"The save() function is generally the right way to update a document with Mongoose. With save(), you get full validation and middleware."

https://mongoosejs.com/docs/documents.html#updating-using-queries

## An example

if i have the model

```js
const blogSchema = new mongoose.Schema({
  title: { type: String, required: true },
  author: String,
  url: { type: String, required: true },
  likes: { type: Number, default: 0 },
})
```

where title is required

and my controller is as such

```js
router.put('/:id', async (req, res) => {
  const { title, author, url, likes } = req.body
  const id = req.params.id
  const updateData = { author, url, likes, title }
  const updatedBlog = await BlogModel.findByIdAndUpdate(id, updateData, {
    runValidators: true,
    new: true,
    context: 'query',
  })
  res.status(200).json(updatedBlog)
})
```

and my blog looks like so before the update

```js
{
    "title": "lotr",
     "author": "jrr tolkien",
     "url": "www.rivendale",
     "likes": "1000000"
}
```

and i send the following data through a put request, with no title.

```js
{
     "author": "jrr hartley",
     "url": "www.flyfishing",
     "likes": "1000000"
}
```

afterwards my data will be like so

```js
{
    "title": "lotr",
     "author": "jrr hartley",
     "url": "www.flyfishing",
     "likes": "1000000"
}
```

and jrr hartley didnt write lord of the rings.

vejol:
Thank you for the issue. I have now updated the course material so that the PUT method is implemented using the findByID and save() methods as documentation suggests. The validation should now work as expected. This has been fixed in commit 6b91b0387c855cfb7ecf2bbb961f7f794d44e426.

PR #4148, opened 2025-09-14
Include alternative update method example in part3c.md

J4DER4:


vejol:
Thank you for the pull request! The material previously used the `findByIdAndUpdate` method, but this later caused issues with validations. According to the [Mongoose documentation](https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/issues/3263), the `save()` method is generally the correct way to update a document in Mongoose, as it provides full support for validations. Therefore, I am not accepting this pull request at this time. 

Here is a link to the original issue: https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/issues/3263

PR #4157, opened 2025-09-27
Add GitHub link to example applications reference

aakku106:
The course material mentions that the code for the example applications can be found on GitHub, but no link was provided.   I couldn’t find a direct repo or folder that contains all of the example applications in one place, but I added a link to the official course repository (https://github.com/fullstack-hy2020/fullstack-hy2020.github.io) so that readers have at least one starting point to navigate from.


vejol:
Thank you for the pull request. Each repository is linked separately in the course material at the points where that specific project is used. I don't think it makes sense to provide a link to the general course repository here, as it might give the impression that students should search for the required repositories themselves. Therefore, I will not accept this pull request.

aakku106:
make sense

PR #4158, opened 2025-09-29
Update part3d.md

mkboop:
Fly.io unable to connect to Atlas MongoDB unless you force IPv4 connection. Provided notice of issue and code that fixes the issue.


vejol:
Thank you for the pull request! I fixed this in a separate commit (3d54771) and updated the course material so that IPv4 is always used when connecting to MongoDB Atlas, and added a brief explanation. I believe it is better to provide the necessary information directly in the material rather than in the assignment. Thanks again for bringing this up!

PR #4163, opened 2025-10-12
Rename 'userId' to 'user' in note schema and docs

brendanharrington:


mluukkai:
this can not be changed just in one plave

PR #4171, opened 2025-10-28
Fix: "Part 6d" - query key type mismatch in newNoteMutation cache update

ebubecodes:
## Problem
The `newNoteMutation`'s `onSuccess` handler uses string format for query keys (`'notes'`), but the `useQuery` hook uses array format (`['notes']`). In TanStack Query v4+, these are treated as completely different cache entries, causing the UI to not update after adding a new note.

## Changes
- Updated `queryClient.getQueryData('notes')` → `queryClient.getQueryData(['notes'])`
- Updated `queryClient.setQueryData('notes', ...)` → `queryClient.setQueryData(['notes'], ...)`

## Result
The mutation now correctly updates the same cache entry that `useQuery` is watching, enabling instant UI updates without triggering an extra GET request (as intended by the optimization in Part 6d).

## Related
This fixes a bug in the course material implementation from 'Part 6d: React Query, useReducer and the context', "Optimizing the performance" section.

### DIFF
--- a/src/content/6/fi/osa6d.md
+++ b/src/content/6/fi/osa6d.md
@@ -380,8 +380,8 @@ const App = () => {
     mutationFn: createNote,
     // highlight-start
     onSuccess: (newNote) => {
-      const notes = queryClient.getQueryData('notes')
-      queryClient.setQueryData('notes', notes.concat(newNote))
+      const notes = queryClient.getQueryData(['notes'])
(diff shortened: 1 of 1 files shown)


ebubecodes:
I pushed a follow-up commit to this branch that addresses the same issue, but in the Finnish translation. No other semantic changes. Thanks!"

PR #4176, opened 2025-11-05
Update explanation of React state updates

PeterKementzey:
Clarify timing of state updates in React. I was not sure based on the text if calling this function multiple times is a mistake, because if the update could happen before the current component function is finished, then it's a concurrency issue. (The word asynchronous lead me to think that.) Checking the react docs it seems to be the case that this never happens, the updates are done in between the 2 component runs.

### DIFF
--- a/src/content/1/en/part1d.md
+++ b/src/content/1/en/part1d.md
@@ -315,7 +315,7 @@ Even though a new value was set for _left_ by calling _setLeft(left + 1)_, the o
 setTotal(left + right) 
 ```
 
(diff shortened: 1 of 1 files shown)


PeterKementzey:
I see in the contribution guidelines to add a `Part 1` label, however, I don't think I have the rights :/.

vejol:
Thank you for the pull request! The matter is indeed as you wrote, and your clarification to the material is good. 

Thank you also for pointing out that contributors do not have permission to add labels—I’ve been wondering for a long time why no one adds them 😂 I will also update the contribution guidelines to be up to date.

PR #4190, opened 2025-11-18
Fix a small bug in Part5C

danielxfeng:
After [@testing-library/user-event v14](https://github.com/testing-library/user-event/releases/tag/v14.0.0), APIs now return promises. Lots of examples had already been updated, but a few outdated snippets remained. This patch updates the remaining snippets.


vejol:
Thank you for pointing out this bug! I fixed it in a separate commit because the line `const user = userEvent.setup()` is placed at the start of the test in earlier code examples. I corrected it the same way to keep it consistent with the rest of the material. The `async` keyword was also missing.

PR #4193, opened 2025-11-19
Update part 7b: use the useField hook more explicitly in the example

adamfarhadi:
It is a little more clear to the student if the useField hook is used in several examples instead of only for the name.

Both Finnish and English example updated


mluukkai:
cphanged in new material

PR #4194, opened 2025-11-19
Use the findByIdAndUpdate method and avoid nested promises when updating note

AbelGAlem:
This PR updates the note update example in Part 3 to use `findByIdAndUpdate` instead of `findById` + `save`.

- Avoids nested promises and keeps all logic in a single promise chain
- Uses `new: true` so the handler responds with the updated document
- Updates the surrounding text to briefly explain about `findByIdAndUpdate`, `new` 
- Removes the discussion about nested promises and their impact on readability, as the updated example no longer requires nested promises

The behavior of the endpoint remains the same from the user’s perspective: it still returns 404 if no note is found and returns the updated note otherwise.

### DIFF
--- a/src/content/3/en/part3c.md
+++ b/src/content/3/en/part3c.md
@@ -887,49 +887,27 @@ In both of the "successful" cases of deleting a resource, the backend responds w
 Let's implement the functionality to update a single note, allowing the importance of the note to be changed. The note updating is done as follows:
 
 ```js
-app.put('/api/notes/:id', (request, response, next) => {
-  const { content, important } = request.body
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request! The material previously used the `findByIdAndUpdate` method, but this later caused issues with validations. According to the [Mongoose dokumentation](https://mongoosejs.com/docs/documents.html#updating-using-queries), the `save()` method is generally the correct way to update a document in Mongoose, as it provides full support for validations.

I added short explanation to the material about why findByIdAndUpdate is not used.

Here is a link to the original issue: #3263

PR #4203, opened 2025-12-01
Part 5b: Using 'ref' directly is not allowed.

Pruthuvi9:
Using 'ref' directly is not allowed as it it a special prop (along with 'key') in React which cannot be forwarded to the component. See here https://react.dev/warnings/special-props. It should be renamed to something else.

### DIFF
--- a/src/content/5/en/part5b.md
+++ b/src/content/5/en/part5b.md
@@ -374,6 +374,7 @@ const App = () => {
   const noteFormRef = useRef() // highlight-line
 
   const noteForm = () => (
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request! It is true that ref is a special variable, but here it is used precisely for its special [purpose](https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop).

I suspect there may have been some confusion about the React version being used. The material in Part 5 is written using version 19, while examples in earlier parts were made with React 18. I have added brief instructions at the beginning of Part 5b on how to check the correct React version. Hopefully, this will no longer cause any confusion in the future!

If, however, this is about something else, feel free to revisit the issue and then provide more detailed information about the error you encountered.

PR #4208, opened 2025-12-09
Update explanation for auto-reloading with ts-node-dev

denny-ddfe:
Updated description of previous implementation of auto-reloading. The course so far has used the --watch flag, not nodemon


mluukkai:
fixed in new material

PR #4214, opened 2025-12-22
Improve delete note test: assert removal by id (not content)

AbelGAlem:
**Description** 

- The current test verifies deletion by checking that the deleted note’s content is no longer present.
- content is not _guaranteed_ to be _unique_, so the test can produce false positives/negatives if multiple notes share identical content.

### DIFF
--- a/src/content/3/en/part3c.md
+++ b/src/content/3/en/part3c.md
@@ -887,27 +887,49 @@ In both of the "successful" cases of deleting a resource, the backend responds w
 Let's implement the functionality to update a single note, allowing the importance of the note to be changed. The note updating is done as follows:
 
 ```js
-app.put("/api/notes/:id", (request, response, next) => {
-  const {content, important} = request.body
+app.put('/api/notes/:id', (request, response, next) => {
+  const { content, important } = request.body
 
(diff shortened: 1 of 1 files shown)


vejol:
Thank you for the pull request! There are a few issues that should be fixed before it can be accepted.

The `findByIdAndUpdate` method is not suitable for the use case in Part 3. A short note about this was added to Part 3 in November 2025. The Part 3 related changes should be removed from this pull request.

Using the id in the test in Part 4 is a good idea. However, this change should also be applied to the other relevant places. In Part 4b, there also seems to be another spot where the same lines of code are shown.

AbelGAlem:
Made the changes. If there is anything else please let me know.

> Thank you for the pull request! There are a few issues that should be fixed before it can be accepted.
> 
> The `findByIdAndUpdate` method is not suitable for the use case in Part 3. A short note about this was added to Part 3 in November 2025. The Part 3 related changes should be removed from this pull request.
> 
> Using the id in the test in Part 4 is a good idea. However, this change should also be applied to the other relevant places. In Part 4b, there also seems to be another spot where the same lines of code are shown.

PR #4218, opened 2026-01-01
Clarify browser behavior during long loop execution

404zhan:
Updated explanation regarding browser responsiveness during long loops and setTimeout usage.


vejol:
Thanks for the pull request! It’s better to fix incorrect content and bring it up to date rather than just add a note about the current state. I updated this in a separate commit.

PR #4222, opened 2026-01-04
Part13b (ES): fix typos, improve formatting

faisne:
- replaced `<i>` tags with asterisks around commands, table names, etc, for consistency with other parts
- fixed small gramatical or translation errors
- improved formatting in code blocks

### DIFF
--- a/src/content/13/es/part13b.md
+++ b/src/content/13/es/part13b.md
@@ -23,7 +23,7 @@ controllers
   notes.js
 ```
 
-El contenido de los archivos es el siguiente. El archivo <i>util/config.js</i> se encarga de manejar las variables de entorno:
+El contenido de los archivos es el siguiente. El archivo *util/config.js* se encarga de manejar las variables de entorno:
 
 ```js
 require('dotenv').config()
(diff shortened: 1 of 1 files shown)


vejol:
Thanks for the pull request! 

Unfortunately, it contains some unwanted and unnecessary changes, so I won’t approve it as-is.

In the course material, the \<i\> tag is used to produce italic text, for example to emphasize file names and new terms. The symbols * and _ around a word make the word monospace-styled, which works well for anything that refers to code, such as variable and function names. Unfortunately, this styling hasn’t been used completely consistently everywhere in the course material, but we should still aim for it.

Here is an example that shows both styles:

image

It is not appropriate to replace all \<i\> tags with the * symbol, because they serve different purposes. There is also no need to replace _ characters with * symbols, because they do the same thing.

Can I ask you to update this pull request so that it follows these principles? Thank you!

mluukkai:
will be taken into accout in another commit

faisne:
Oh, I'm sorry, I was meaning to fix the PR, but didn't get around to do it. Thanks!

PR #4231, opened 2026-01-14
Fix the CounterContext path to be correct

not6248:


vejol:
Thanks for the pull request! Since the files _CounterContext.jsx_ and _App.jsx_ are both located in the same _/src_ directory, the import used in the course material is correct. Therefore, I won’t be accepting this pull request.

PR #4232, opened 2026-01-16
use setQueryData updater function for notes state update

KiwiUserSFTW:
This change replaces the cache update logic with the updater function of setQueryData, which allows the existing notes state to be updated more concisely. The explanation was adjusted accordingly.

"Maybe the previous version is considered more suitable for teaching purposes, this change can be treated as a light suggestion"

Thank you for a great course


vejol:
Thanks for the pull request! In my opinion, the syntax used in the course material is simpler and easier to read, so I won’t be accepting this change.

PR #4238, opened 2026-01-27
Fix Togglable component by adding forwardRef

IMJUSTWANNADEPLOYABOT:


idealistichacker:
I encountered the same issue while learning Part 5 with React 18 and Vite. This fix is definitely necessary. Thanks for proposing this!

vejol:
Thank you for the pull request! The material in Part 5 is written using React 19, which [no longer uses](https://react.dev/reference/react/forwardRef ) `forwardRef`. 

Therefore, I cannot accept these changes. I have added a note at the beginning of Chapter 5b about this, along with instructions on how to check the correct React version. Hopefully this will no longer cause any confusion in the future!

PR #4258, opened 2026-03-20
Fix ES6 link in osa1a.md

sandraole:
The domain registration of the original website has lapsed and it now points to a random irrelevant website. This commit replaces the link with a link to the official ECMAScript standard.


vejol:
Thanks for the pull request and for pointing this out! I fixed it in a separate commit, as the link you provided actually pointed to the ES16 version. I also updated the link across all language versions.

PR #4284, opened 2026-05-15
docs: fix duplicate code snippet typo in Zustand selector example (Part 6a)

sahaarnav3:
### Description
Fixes a copy-paste snippet error in the Zustand state selection documentation section of `part6a.md`. 

### Context
The explanatory text states: *"When instead writing: ... the component no longer reacts to changes in the counter value"*. However, the code block directly beneath it mistakenly duplicated the original unoptimized destructuring snippet (`const { increment, decrement, zero } = useCounterStore()`) from line 359.

### Changes Made
Updated lines 365–367 to demonstrate the actual atomic selector pattern intended by the text:
```js
const increment = useCounterStore((state) => state.increment)
const decrement = useCounterStore((state) => state.decrement)
const zero = useCounterStore((state) => state.zero)
```


mluukkai:
fixed in another commit

PR #4289, opened 2026-05-23
fix: clarify part3b backend reference to avoid ambiguity with json-server

Dinokojt7:
The opening instruction in part3b says:

"Next, let's connect the frontend we made in part 2 to our own backend."

The phrase "our own backend" is ambiguous. A reader could reasonably interpret it as the Part 2 json-server setup, since that backend was also built progressively through the course exercises and lives in the Part 2 project root — it qualifies just as much as "our own backend."

When a reader makes that assumption and updates baseUrl to http://localhost:3001/api/notes while json-server is still running, they get a 404 instead of the CORS error the section is trying to demonstrate. This is because json-server has no /api/notes route — it only serves /notes. The reader then has no clear signal that they are pointed at the wrong backend entirely.

The fix replaces "our own backend" with "the Express backend we built in part 3a" and updates the surrounding sentence to make the Part 3 context explicit, so the reader does not need to make the cognitive leap between parts themselves.

### DIFF
--- a/src/content/3/en/part3b.md
+++ b/src/content/3/en/part3b.md
@@ -101,6 +101,7 @@ There are also some other free hosting options that work well for this course, a
 Some course participants have also used the following services:
 
 - [Replit](https://replit.com)
+- [Railway](https://railway.app)
 - [CodeSandBox](https://codesandbox.io)
 
 If you know easy-to-use and free services for hosting NodeJS, please let us know!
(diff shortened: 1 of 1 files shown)


Dinokojt7:
Also added Railway to the list of free hosting alternatives alongside Replit and CodeSandBox. It's a purpose-built PaaS with a free tier, no local installations required, and better suited for Node/Express deployment than CodeSandBox. Left both existing options in — happy to remove CodeSandBox if you'd prefer a shorter list.

PR #4307, opened 2026-07-22
Make crosses on info banners visible in dark mode

Sseik:
Crosses on info banners were turning white in the dark mode, which made them barely visible on the light background, so I changed the buttonStyle to always use the color it uses in the light mode. Also, I had to create an empty .cache directory in the root before I could start the app, you might want to add it to the CONTRIBUTING.md


adriamarcet:
I have added [a similar PR](https://github.com/fullstack-hy2020/fullstack-hy2020.github.io/pull/4323) but only for Part 6 infoBanner.

PR #4309, opened 2026-07-23
Update part3c.md - A true fullstack developer's oath

UnFoxable:


UnFoxable:
Change the grammer abit
I will keep an eye on the database: does the backend save data there in the right format
-> I will keep an eye on the database; whether the data is saved in the expected state

PR #4316, opened 2026-08-12
Fix typo in part7b.md regarding transpilation

chunogreg:
const element = <App />copy
it must be **transpiled it into** something the browser can run:

const element = React.createElement(App, null)

correction:  it must be **transpiled into** something the browser can run:


chunogreg:
removed 'it' from:  transpiled it into

